PR #1, opened 2025-05-12
Tweaks for more runtime environments

bunnie:
This PR introduces two features

1. Adds a convenience argument `--clear-caches ` that removes all the staging files starting with .~. This is helpful when debugging partially aborted runs.
2. Adds a retry on the portion of the script that moves the staging directories into output directories. On Windows, the antivirus will lock the directories shortly after they are populated, which can cause them to be inaccessible. This patch catches failures due to this and retries the move with a short pause to allow the AV to finish checking the directory so that it can me moved to its final location.


jadephilipoom:
Sorry for the delay, just now saw this PR when looking through the PR list. Looks fine to me, but @matutem is more of an authority on the ipgen/topgen tooling.

PR #8, opened 2025-05-29
RFC: Initial explorations in PQC integration for OTBN

bunnie:
The intent of this PR is an RFQ from maintainers on methodology and approach to PQC integration for OTBN.

This is the basic strategy:
- Pull OTBN simulation code from https://github.com/PQC-OpenTitan/towards-ml-kem-and-ml-dsa-on-opentitan-dev-multi-cycle-bnmulv
- Pull a single test case from the above repo
- Try to get the test case to pass

In trying to get this to pass, some problems were encountered in the expected format of the `.exp` file for `dmem`. These issues are noted in the commit comments, and will be highlighted as comments in the code below as well.

### DIFF
--- a/hw/ip/otbn/data/otbn.hjson
+++ b/hw/ip/otbn/data/otbn.hjson
@@ -34,7 +34,7 @@
           notes:              ""
       },
       {
-          version:            "1.0.0",
+          version:            "1.1.0",
           life_stage:         "L1",
           design_stage:       "D2S",
           verification_stage: "V2S",
@@ -757,7 +757,7 @@
     // Imem size (given as `items` below) must be a power of two.
     { window: {
         name: "IMEM",
-        items: "8192", // 4 kB
+        items: "8192", // 32 kB
         swaccess: "rw",
         data-intg-passthru: "true",
         byte-write: "false",
@@ -781,7 +781,6 @@
     { window: {
         name: "DMEM",
         items: "32768", // 32768 * 32bit = 128KiB, all visible over bus
-        unusual: "true", // Needed to avoid an error because of the non-power-of-two size
         swaccess: "rw",
         data-intg-passthru: "true",
         byte-write: "false",
--- a/hw/ip/otbn/dv/otbnsim/sim/csr.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/csr.py
@@ -1,4 +1,4 @@
-# Copyright lowRISC contributors.
+# Copyright lowRISC contributors (OpenTitan project).
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 # Modified by Authors of "Towards ML-KEM & ML-DSA on OpenTitan" (https://eprint.iacr.org/2024/1192).
--- a/hw/ip/otbn/dv/otbnsim/sim/dmem.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/dmem.py
@@ -55,7 +55,7 @@ class Dmem:
         # if the word has invalid integrity bits and we'll get an error if we
         # try to read it. Otherwise, we store the integer value.
         num_words = dmem_size // 4
-        self.data = [None] * num_words  # type: List[Optional[int]]
+        self.data: List[Optional[int]] = [None] * num_words
 
         # Because it's an actual memory, stores to DMEM take two cycles in the
         # RTL. We wouldn't need to model this except that a DMEM invalidation
@@ -66,8 +66,8 @@ class Dmem:
         # this cycle. However, the first commit() will then move it to the
         # self.pending list. Entries here will only make it to self.data on the
         # next commit().
-        self.trace = []  # type: List[TraceDmemStore]
-        self.pending = {}  # type: Dict[int, int]
+        self.trace: List[TraceDmemStore] = []
+        self.pending: Dict[int, int] = {}
 
     def _load_5byte_le_words(self, data: bytes) -> None:
         '''Replace the start of memory with data
--- a/hw/ip/otbn/dv/otbnsim/sim/edn_client.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/edn_client.py
@@ -18,12 +18,12 @@ class EdnClient:
         # non-None if we're in the middle of reading a value from the EDN. This
         # should always have length <= ACC_LEN (sending an extra beat of data
         # causes an error).
-        self._acc = None  # type: Optional[List[int]]
+        self._acc: Optional[List[int]] = None
 
         # A counter of the number of beats we've been waiting since self._acc
         # became full before being told that CDC was done. This is None unless
         # self._acc contains a list of ACC_LEN values.
-        self._cdc_counter = None  # type: Optional[int]
+        self._cdc_counter: Optional[int] = None
 
         # If true, the next transaction that completes will be discarded.
         self._poisoned = False
@@ -35,7 +35,7 @@ class EdnClient:
         self._fips_err = False
         self._rep_err = False
 
-        self._last_word = None  # type: Optional[int]
+        self._last_word: Optional[int] = None
 
     def request(self) -> None:
         '''Start a request if there isn't one pending'''
--- a/hw/ip/otbn/dv/otbnsim/sim/ext_regs.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/ext_regs.py
@@ -129,8 +129,8 @@ class RGReg:
     def __init__(self, fields: List[RGField], double_flopped: bool):
         self.fields = fields
         self.double_flopped = double_flopped
-        self._changes = []  # type: List[ExtRegChange]
-        self._next_changes = []  # type: List[ExtRegChange]
+        self._changes: List[ExtRegChange] = []
+        self._next_changes: List[ExtRegChange] = []
 
     @staticmethod
     def from_register(reg: Register, double_flopped: bool) -> 'RGReg':
@@ -146,7 +146,10 @@ class RGReg:
             new_val |= field_new_val << field.lsb
         return new_val
 
-    def write(self, value: int, from_hw: bool) -> None:
+    def write(self,
+              value: int,
+              from_hw: bool,
+              immediately: bool = False) -> None:
         '''Stage the effects of writing a value.
 
         If from_hw is true, this write is from OTBN hardware (rather than the
@@ -155,7 +158,8 @@ class RGReg:
         '''
         assert value >= 0
         now = self._apply_fields(lambda fld, fv: fld.write(fv, from_hw), value)
-        changes = self._next_changes if self.double_flopped else self._changes
+        delayed = self.double_flopped and not immediately
+        changes = self._next_changes if delayed else self._changes
         changes.append(ExtRegChange('=', value, from_hw, now))
 
     def set_bits(self, value: int) -> None:
@@ -250,7 +254,7 @@ class OTBNExtRegs:
     def __init__(self) -> None:
         _, reg_block = load_registers()
 
-        self.regs = {}  # type: Dict[str, RGReg]
+        self.regs: Dict[str, RGReg] = {}
         self._dirty = 0
 
         assert isinstance(reg_block, RegBlock)
@@ -287,10 +291,14 @@ class OTBNExtRegs:
             raise ValueError('Unknown register name: {!r}.'.format(reg_name))
         return reg
 
-    def write(self, reg_name: str, value: int, from_hw: bool) -> None:
+    def write(self,
+              reg_name: str,
+              value: int,
+              from_hw: bool,
+              immediately: bool = False) -> None:
         '''Stage the effects of writing a value to a register'''
         assert value >= 0
-        self._get_reg(reg_name).write(value, from_hw)
+        self._get_reg(reg_name).write(value, from_hw, immediately)
         self._dirty = 2
 
     def set_bits(self, reg_name: str, value: int) -> None:
--- a/hw/ip/otbn/dv/otbnsim/sim/flags.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/flags.py
@@ -38,7 +38,7 @@ class FlagReg:
         self.L = L
         self.Z = Z
 
-        self._new_val = None  # type: Optional['FlagReg']
+        self._new_val: Optional['FlagReg'] = None
 
     def set_flags(self, other: 'FlagReg') -> None:
         self._new_val = other
--- a/hw/ip/otbn/dv/otbnsim/sim/insn.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/insn.py
@@ -331,7 +331,7 @@ class LW(OTBNInsn):
         base = state.gprs.get_reg(self.grs1).read_unsigned()
         if state.gprs.call_stack_err:
             state.stop_at_end_of_cycle(ErrBits.CALL_STACK)
-            return
+            return None
 
         addr = (base + self.offset) & ((1 << 32) - 1)
 
@@ -339,7 +339,7 @@ class LW(OTBNInsn):
             if DEBUG_MEM:
                 print(f"lw {base} {self.offset}: failed", file=sys.stderr)
             state.stop_at_end_of_cycle(ErrBits.BAD_DATA_ADDR)
-            return
+            return None
 
         result = state.dmem.load_u32(addr)
 
@@ -351,12 +351,13 @@ class LW(OTBNInsn):
 
         if result is None:
             state.stop_at_end_of_cycle(ErrBits.DMEM_INTG_VIOLATION)
-            return
+            return None
 
         if DEBUG_MEM:
             print(f"\t{format(result, '08x')}", file=sys.stderr)
 
         state.gprs.get_reg(self.grd).write_unsigned(result)
+        return None
 
 
 class SW(OTBNInsn):
@@ -522,12 +523,12 @@ class CSRRS(OTBNInsn):
         if not state.csrs.check_idx(self.csr):
             # Invalid CSR index. Stop with an illegal instruction error.
             state.stop_at_end_of_cycle(ErrBits.ILLEGAL_INSN)
-            return
+            return None
 
         bits_to_set = state.gprs.get_reg(self.grs1).read_unsigned()
         if state.gprs.call_stack_err:
             state.stop_at_end_of_cycle(ErrBits.CALL_STACK)
-            return
+            return None
 
         if self.csr == 0xfc0:
             # A read from RND. If a RND value is not available, request_value()
@@ -544,6 +545,8 @@ class CSRRS(OTBNInsn):
         if self.grs1 != 0:
             state.write_csr(self.csr, new_val)
 
+        return None
+
 
 class CSRRW(OTBNInsn):
     insn = insn_for_mnemonic('csrrw', 3)
@@ -559,12 +562,12 @@ class CSRRW(OTBNInsn):
         if not state.csrs.check_idx(self.csr):
             # Invalid CSR index. Stop with an illegal instruction error.
             state.stop_at_end_of_cycle(ErrBits.ILLEGAL_INSN)
-            return
+            return None
 
         new_val = state.gprs.get_reg(self.grs1).read_unsigned()
         if state.gprs.call_stack_err:
             state.stop_at_end_of_cycle(ErrBits.CALL_STACK)
-            return
+            return None
 
         if self.csr == 0xfc0 and self.grd != 0:
             # A read from RND. If a RND value is not available, request_value()
@@ -582,6 +585,7 @@ class CSRRW(OTBNInsn):
             state.gprs.get_reg(self.grd).write_unsigned(old_val)
 
         state.write_csr(self.csr, new_val)
+        return None
 
 
 class ECALL(OTBNInsn):
@@ -1362,7 +1366,7 @@ class BNLID(OTBNInsn):
 
         if self.grs1_inc and self.grd_inc:
             state.stop_at_end_of_cycle(ErrBits.ILLEGAL_INSN)
-            return
+            return None
 
         grs1_val = state.gprs.get_reg(self.grs1).read_unsigned()
         addr = (grs1_val + self.offset) & ((1 << 32) - 1)
@@ -1387,7 +1391,7 @@ class BNLID(OTBNInsn):
             saw_err = True
 
         if saw_err:
-            return
+            return None
 
         wrd = grd_val & 0x1f
         value = state.dmem.load_u256(addr)
@@ -1405,10 +1409,13 @@ class BNLID(OTBNInsn):
 
         if value is None:
             state.stop_at_end_of_cycle(ErrBits.DMEM_INTG_VIOLATION)
-            return
+            return None
+
         if DEBUG_MEM:
             print(f"\t {format(value, '064x')}", file=sys.stderr)
+
         state.wdrs.get_reg(wrd).write_unsigned(value)
+        return None
 
 
 class BNSID(OTBNInsn):
@@ -1425,7 +1432,7 @@ class BNSID(OTBNInsn):
     def execute(self, state: OTBNState) -> Optional[Iterator[None]]:
         if self.grs1_inc and self.grs2_inc:
             state.stop_at_end_of_cycle(ErrBits.ILLEGAL_INSN)
-            return
+            return None
 
         grs1_val = state.gprs.get_reg(self.grs1).read_unsigned()
         addr = (grs1_val + self.offset) & ((1 << 32) - 1)
@@ -1450,7 +1457,7 @@ class BNSID(OTBNInsn):
             saw_err = True
 
         if saw_err:
-            return
+            return None
 
         if self.grs1_inc:
             new_grs1_val = (grs1_val + 32) & ((1 << 32) - 1)
@@ -1467,6 +1474,7 @@ class BNSID(OTBNInsn):
         if DEBUG_MEM:
             print(f"\t {format(wrs_val, '064x')}", file=sys.stderr)
         state.dmem.store_u256(addr, wrs_val)
+        return None
 
 
 class BNMOV(OTBNInsn):
@@ -1497,7 +1505,7 @@ class BNMOVR(OTBNInsn):
             eprint("MOVR")
         if self.grs_inc and self.grd_inc:
             state.stop_at_end_of_cycle(ErrBits.ILLEGAL_INSN)
-            return
+            return None
 
         grd_val = state.gprs.get_reg(self.grd).read_unsigned()
         grs_val = state.gprs.get_reg(self.grs).read_unsigned()
@@ -1520,7 +1528,7 @@ class BNMOVR(OTBNInsn):
             saw_err = True
 
         if saw_err:
-            return
+            return None
 
         wrd = grd_val & 0x1f
         wrs = grs_val & 0x1f
@@ -1537,6 +1545,7 @@ class BNMOVR(OTBNInsn):
 
         value = state.wdrs.get_reg(wrs).read_unsigned()
         state.wdrs.get_reg(wrd).write_unsigned(value)
+        return None
 
 
 class BNWSRR(OTBNInsn):
@@ -1552,7 +1561,7 @@ class BNWSRR(OTBNInsn):
         if not state.wsrs.check_idx(self.wsr):
             # Invalid WSR index. Stop with an illegal instruction error.
             state.stop_at_end_of_cycle(ErrBits.ILLEGAL_INSN)
-            return
+            return None
 
         if self.wsr == 0x1:
             # A read from RND. If a RND value is not available, request_value()
@@ -1575,13 +1584,14 @@ class BNWSRR(OTBNInsn):
         # provided us with a value). If not, fail with a KEY_INVALID error.
         if not state.wsrs.has_value_at_idx(self.wsr):
             state.stop_at_end_of_cycle(ErrBits.KEY_INVALID)
-            return
+            return None
 
         # The WSR is ready and has a value. Read it.
         val = state.wsrs.read_at_idx(self.wsr)
         if DEBUG_KMAC:
             eprint(f"read WSR: {format(val, '064x')}")
         state.wdrs.get_reg(self.wrd).write_unsigned(val)
+        return None
 
 
 class BNWSRW(OTBNInsn):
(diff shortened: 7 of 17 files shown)


hw/ip/otbn/util/otbn_sim_test.py
```diff
@@ -152,20 +152,25 @@ def main() -> int:
...
-                    if actual_dmem[offset:offset + len(value)] != value:
-                        result.err(
-                            f"Mismatch for dmem {label}:\n"
-                            f"  Expected: {value.hex()}\n"
-                            f"  Actual:   {actual_dmem[offset:offset+len(value)].hex()}"
-                        )
-                except KeyError:
-                    result.err(f'No label "{label}" found in elf-file.')
+            for i in range(0, len(expected_dmem)):
+                if expected_dmem[i] is not None and expected_dmem[i] != actual_dmem[i]:
```
bunnie, line 156: This is the way dmem memory comparisons work in the PQC-opentitan fork. This is quite inconsistent with the existing code which is retained as commented material below.

My guess is this will almost certainly break some other tests elsewhere that are relying upon annotated ELF files for dmem comparisons. So a decision needs to be made if the idea is to retrofit all the existing tests from PQC-opentitan to use this other format, or if we create perhaps a third format altogether?

hw/ip/otbn/util/shared/dmem_dump.py
```diff
@@ -47,9 +59,11 @@ def parse_actual_dmem(dump: bytes) -> bytes:
     Returns the dmem bytes except integrity info.
     '''
     dmem_bytes = []
+    hexdump = dump[6:]
+    bindump =  bytes.fromhex(hexdump.decode('ascii'))
```
bunnie, line 63: Note that this crazy bit of code is because the data as passed to us is not actually `bytes` but hex encoded in ASCII. The actual format is something like
```
dmem:
23abcd...
```
i.e. the first six bytes are literally `b'dmem:\n'` and thus must be skipped, and the remainder is an ASCII string that has to be converted to a number before turning into a bytes object. This feels like there is a mis-alignment somewhere in the APIs but I don't have the requisite knowledge to know where that's happening. This hack allows the singular test case I have run to pass, though, so it is at least points to a way forward?

hw/ip/otbn/util/otbn_sim_test.py
jadephilipoom: Most tests currently in the repo just use registers for expected values, because the DMEM parsing was added later. In fact, I see only 2 tests with a `.dexp` file. So the path of least resistance would probably be to modify those two tests and leave all the PQC tests as-is.

However, I think the version currently in expo is more usable long-term, because it allows listing expected values in terms of _labels_ rather than explicit DMEM offsets. If you have multiple DMEM labels across multiple source files, it's not always clear where those buffers are going to land relative to each other in DMEM at link time. Looking at the PR where DMEM expected-value parsing was merged upstream, it looks like the PQ authors made this change before submitting; it's not discussed in review. This implies they also thought it was preferable, or at least that there wasn't a strong reason to stick with explicit offsets. So I do have some level of preference for keeping the current expo version.

An in-between option would be to adjust `otbn_sim_test.py` to accept _both_ formats for now, finish adding the PQ tests, and then go through later to make the format consistent. This maybe seems like the best way to separate concerns and deal with the merge/update separately from this particular disagreement. What do you think?

jadephilipoom: I guess the other possibility is to change or disable the two tests for now, and only accept this format for the time being while we merge in code.

bunnie: I agree that the the labeled test values are more sane than explicit pokes, and more long-term sustainable.

I think what I'd need to do is then to write a script that can convert the existing .exp files to .dexp files, and back out these two patches to POC the integration. If you think that's a reasonable approach, I can give it a whack.

I think it'd be a one-time use script so maybe I would not even check it in to the repo, I'd just run it, convert the format from the PQC-opentitan, and then after conversion commit them into the repo.

Of course, this means that if there is a bug in my script it's a little harder to check because you lose the history of the original test - I suppose we could check in both the deprecated .exp file and the .dexp file, and then remove it from the commit history just so that it's somewhere in the logs in case we need to refer to it again?

jadephilipoom: > I think what I'd need to do is then to write a script that can convert the existing .exp files to .dexp files, and back out these two patches to POC the integration. If you think that's a reasonable approach, I can give it a whack.

So you mean the script would convert to the new format, presumably by finding contiguous DMEM sequences in the .exp file and then looking for labels in the binary at the start offset? That does sound reasonable, it would limit the short-term pain of converting formats. I'd be in favor of checking in the script itself so that we have a record of it, and we can always delete it later. Checking in the .exp files and then converting them as a separate commit feels reasonable too.

bunnie: Yes, that's basically the idea of the script. I'd probably "test" it in the process of shuffling the existing PQ tests into the final locations in this PR, under the presumption that the tests passed and any failures are due to bugs in the script or something else going on.

I think because the file name extensions are different having the .exp and .dexp files concurrently is not unreasonable. We can just accumulate them in this RFC PR and then remove them before the merge, I guess? so at the very least if we had to puzzle through a bug they'd be referenced in this PR and a commit tag.

bunnie:
@jadephilipoom please see the comments in commit https://github.com/zerorisc/expo/pull/8/commits/c18853996cf971edc0bb9e56d166987d2cf8a62b for details on the proposed conversion tool for expected test data formats coming from the pqc-opentitan branch.

I have added another test case, poly_pointwise_dilithium. I'm pretty sure the .exp file was correctly converted over (just by eyeballing it) but the test case itself actually fails - the expected data does not match the actual output. I think this might indicate that blindly copying over the otbnsim artifacts from the pqc-opentitan branch does a complete job? Not sure.

Note also that currently this PR does not have the changes in otbn_as.py or otbn_ld.py that are in pqc-opentitan, but before trying to conflate all that in, wanted to get your feedback on the test case conversion script before diving into debugging the other problems we are seeing.

jadephilipoom:
I took a closer look at this just now. I was able to replicate the failing test both on a local copy of this branch _and_ on a fresh clone of the source repo for the PQ work. I think I figured out what's happening. The test is setting the MOD register simply to 0x7fe001, which is indeed the Dilithium modulus. But the simulator is actually reading two values from there:
https://github.com/PQC-OpenTitan/towards-ml-kem-and-ml-dsa-on-opentitan-dev-multi-cycle-bnmulv/blob/45b718bab9a9bef9d09dfddef9340337d99942e8/hw/ip/otbn/dv/otbnsim/sim/insn.py#L804

It's reading the modulus from the lowest part of the word, but then it's also looking for the Montgomery constant related to the modulus in a different register. This value is then reading as zero, which means all the products come out as very small numbers. The top-level tests seem to indeed provide the modulus in this format, e.g.:
https://github.com/PQC-OpenTitan/towards-ml-kem-and-ml-dsa-on-opentitan-dev-multi-cycle-bnmulv/blob/45b718bab9a9bef9d09dfddef9340337d99942e8/sw/otbn/crypto/tests/verify_dilithium_test.s#L106

However, the smaller unit tests don't. I think the design probably didn't originally take the Montgomery constant from MOD, and when this update went through only the top-level tests got updated. So they're all failing on the source repo!

I changed the test to load the Montgomery constant as well, and then I got correctly-sized numbers but the test still didn't pass. Turning on the arithmetic debugging in `hw/ip/otbn/dv/otbnsim/insn.py`, I could see things like:

bunnie:
I don't know the build system well enough to answer this myself, so maybe this sounds like a dumb question - but is there a CI bench for the PQC implementation that gets run regularly? I would presume that would contain the list of actively maintained tests, and that would serve as guidance as to which tests to focus on porting first.

bunnie:
Most recent commit set adds the ability to pass a `-D` to the test code via a `-copt` argument to bazel. Also overwrites otbn_as and otbn_ld with the version from pqc-opentitan just to see if this makes the tests run more consistently with what is over there.

Finally an mlkem test is emplaced which can run. I am using `./bazelisk.sh test --copt="-DKYBER_K=3" --sandbox_debug //sw/otbn/crypto/tests/mlkem:kyber768_mlkem_keypair_test` and the test fails. Here's the result:

```
ERROR: Mismatch for dmem stack_end:
  Expected: 1c9dc524b7e703761baac74b213acbc2ebfa3c4b5bd83bb62784406539ba98c427bb7143b5a3921f541db8060d7c5aa951a1badff3d653951b605acde9b0b67d4649511aad681f8ff38b4726090e67c69ec4c43a46ba907a4ce95d592941e05041a811a8a83495b3b1e70a7fe25365116b569a0cc7f78f9bb2b828e703348901a552f2a44c87305278896b2549a3cd34cb257b8086305ad73280fb7b7f010082560bb71c656b54ccdb9cdcd3f17c10459634ba0dc33a041946950b5c90239a30d53e81393a350fc419428e5e614964356ca5bd121d088cb39cae52f21a447e2ca7922e06247adac8d852990cb61b5f6ba5383ba5a8540aac2df1434b5652d0a1124a6855b2600b098d620c8b550192b00a07d1472e19d6f56136966354c6035d8c0090bb41785ba18a17f619f4bed7004a274aa55ce522c984a8a3616358a20abca384946c3be4e227119b96a1da3156a01ea629089f93e210a1e177244ba4c807bb4fc59fdb58a9b51ecaee876c082bb0a943bf472c5b2a3a7c11624e4cae99337aaa2e7314a7b903c115726c4f511792ef9942484cd6ace75ce8588901a83738d722002340351ff58e0c0c288a840abfa09bb09c61508b605157907629761dc0c9497421e3bab27fa738f4e4552a41b3b4ca5e56c6539039c60118be9564beef44a18f66af296091ca0734e56d948123c7aef33bab4363098aa3c52b411bc023fb0aaff2b8e9f9f21008b4cdfb4fce7279d8bf06983e32babc6051d6afb33447665bc29aa9fc45ef5ca79e84769d013323ec85e4399b6d588d5dd03d63307be5ad694f68393a754fe43502a95d9927a8f398679f4a185e083219ad419f6229b77b41401a85f5db135892b788d04641997427451c7a081a89eaea2b7b0818f225ae74a7274c493be01d0b7553d9444bb17ae4c55e20650fc3f912e229e59103499553950b3a871643451641b0c001e8f444cb9958679589400bc12391acbbf4ca58a5775c76300c6afdab2a844ca54d0239e31343063fd64946f030e29d42823bd04c609f4ae84d70541a82a9403a4218c4dc64773d366327312cb02e944f30c6981ce75327a4431000f33e25b84849aa5379275585fdab775af78813a299d1168134adf639e5e9a56870d35410c7517e2c7a8bd99c5fe9323348b34bc43c6778a718cbc75594a440b3e068ac0211f864b0766bef51710543e11a1085547f78f88d5553e03c5a03cd439c0ccc4231a4b03490a389607e9a644c7b86ec690a9818682eabcb354c72455d9d909c346abc5606b371c0ad647698f68ecc4b4b9b04f6d02d435807ac9ff7c4770f90bb229b41b397cb851607884c69313472f62091578dd54ad9ce7f122962ff99c2c3d4a5c89c124036c7ac12d849f998f26136e7bf628252df5cfacc6404215a95e20698171216eaaa653308aee6ce26e856132ffccec5e1440357c80a16f17e8c0e4a1b38177b01886cc6029ad370ab56c1966c24865a9d08eb7e5a1922687d7468070824bf445b67b5654fba8d65b2a52b96e61821322badcf14e20235cfd7cac92d58a952a9fb60d37a6397251007239dd978a95d23f84b609107764f0a4f590f68403dd933b62b2884fb4b674ca5a4845611b631c31b041aa6595e0c3d882777c0bb3bc3a4542ee5906a06ce7806a83b45a05ab188a16aae6b49637b1724b0fb126d95cdf2880220ac0431304731625c1aa3fb8862e48b85db3ea50ebc21d12f219a9ea0dbc6d01207927abea268a3f2f4bcee43345341dd801760a981965faf90cd158cabca5725ef522b7a152f184a93bb767ac5a55da4541bbcc64a77a1833559cd4dfcbcab87ca0c3d46ff9cbb0ee8c58cd101b524e3398ba07cd0dbba75bf6cbc7a2933d5bdaac71b4008b333f587a3517b92f3f58013e99e119ba5ab3548352cb62d185fd8d676744132a62a01523a221228ace23700bc77b493c459cb997a55a1ab299bbc06a05f361c46ac3f58a303b41454073591af0da1339bb49c81f82567f062c4b3923fc1b17f5a91408d47ba1090e31910074415aa0d01143df4c0a38f1a64a2fa4ca981ae77a822f8bb132332e11efb94643759828b125a10d552d67d15e4d5710fd74df348bd62769946126cce9bca881c935e968208278f53276b79192b07261225381b135b15904f569ccd83958c093c4c67a2068f73b862b229969e1b5211c476b5306ca75642d002437b3535ce3e3acd1099bcf59eca89bbe875ba91c2531d1222529b320d5612df522427246e0bf63f990d31ad3b484d95723a38911b7f936ab274b46019b8a52e39035854479f1cd83f4ab4baa733c8e0b3f5a1db452a5a3a5c64129831c237651bd76371a5a4d77ba1289a45aa5986c29acaa1d90663a3995533452da6458410778e437375734e62d0841a7cb20fcf31aa87114b8e73e6fbb4f5ba9a1f4b002b2664d9a04232c5d31e87dd354d36897a67bb5e338109ad18148b2054eb432b3b81d85f9ed425526c0908b494039945b21d6a48c61434acb967185e942f9c86b5dc9ecf888166210ad3784580a3e52349a9bbba537790db23e784b3257298664ee007e092f266675cb739f867e2557c6112a20f8b37d0a21df90a83d6fbe169d864cbb052a721e25fe4462a9b46d380198dc5cc714b055687137bca2cf3055a71cc5553928dab38a85f25252427571361bcd67a16bf2e63b0954b38868810d20aa4b492c29ce5c0fc7ff7d3429fcd85acb349b05ca90595a2acc6c4840a3dca98b1d2308fb91598bd1b05274b45c3ff409e8b665a612814141b0c281bf6552b8f0ce104eb0e34f168152158f1b7e20ef1b0ebb785610c3aec81f40c67cd94b5070d8e236e471b966887873b989d51719768217c2bb494d41c003df020eb6eb04b80c0f954f6aff2c2d57490751cb1a3b36daff173ecee74c809544d250fa9eafcb6906df60f540a9318bed9d1861daf44616af91f6c4a411ab28fa76138748ea7bc5eb03a815a85734272628ae2a323d528b45aa364254a5baa7267f7f8c9cdf8b4f1657c059981b5a59e4ffbbdfb6691eec501b241f70dfcb511a28c11a579b02944b9b87f537b84572dd8bee8632d632d21d815430d288aa604863288c1d2c461b07a7e16a020f173f4b28d93b16953eb0a0f7865f21f38e2463b3f7ae7af06774c817e66b1a116a0d27d72aed8a7b9da25e4ac817f97639f7c450376268a439c3a0ec158850b63e58e2873c305ca609e32c77cfa2c509e42b918c95d444405f593fa4cc722e96e7b9360d607c337598f96624d11126d00689602c6dee553592d4c80f318365caddfc137018513f312e5fd70b64764e7cfb5b5c4f63c558300a9eeb13c934b00880afb3c10682afdee4afa016e63a3e858e3a1a89ce257aef9b631c31b041aa6595e0c3d882777c0bb3bc3a4542ee5906a06ce7806a83b45a05ab188a16aae6b49637b1724b0fb126d95cdf2880220ac0431304731625c1aa3fb8862e48b85db3ea50ebc21d12f219a9ea0dbc6d01207927abea268a3f2f4bcee43345341dd801760a981965faf90cd158cabca5725ef522b7a152f184a93bb767ac5a55da4541bbcc64a77a1833559cd4dfcbcab87ca0c3d46ff9cbb0ee8c58cd101b524e3398ba07cd0dbba75bf6cbc7a2933d5bdaac71b4008b333f587a3517b92f3f58013e99e119ba5ab3548352cb62d185fd8d676744132a62a01523a221228ace23700bc77b493c459cb997a55a1ab299bbc06a05f361c46ac3f58a303b41454073591af0da1339bb49c81f82567f062c4b3923fc1b17f5a91408d47ba1090e31910074415aa0d01143df4c0a38f1a64a2fa4ca981ae77a822f8bb132332e11efb94643759828b125a10d552d67d15e4d5710fd74df348bd62769946126cce9bca881c935e968208278f53276b79192b07261225381b135b15904f569ccd83958c093c4c67a2068f73b862b229969e1b5211c476b5306ca75642d002437b3535ce3e3acd1099bcf59eca89bbe875ba91c2531d1222529b320d5612df522427246e0bf63f990d31ad3b484d95723a38911b7f936ab274b46019b8a52e39035854479f1cd83f4ab4baa733c8e0b3f5a1db452a5a3a5c64129831c237651bd76371a5a4d77ba1289a45aa5986c29acaa1d90663a3995533452da6458410778e437375734e62d0841a7cb20fcf31aa87114b8e73e6fbb4f5ba9a1f4b002b2664d9a04232c5d31e87dd354d36897a67bb5e338109ad18148b2054eb432b3b81d85f9ed425526c0908b494039945b21d6a48c61434acb967185e942f9c86b5dc9ecf888166210ad3784580a3e52349a9bbba537790db23e784b3257298664ee007e092f266675cb739f867e2557c6112a20f8b37d0a21df90a83d6fbe169d864cbb052a721e25fe4462a9b46d380198dc5cc714b055687137bca2cf3055a71cc5553928dab38a85f25252427571361bcd67a16bf2e63b0954b38868810d20aa4b492c29ce5c0fc7ff7d3429fcd85acb349b05ca90595a2acc6c4840a3dca98b1d2308fb91598bd1b05274b45c3ff409e8b665a612814141b0c281bf6552b8f0ce104eb0e34f168152158f1b7e20ef1b0ebb785610c3aec81f40c67cd94b5070d8e236e471b966887873b989d51719768217c2bb494d41c003df020eb6eb04b80c0f954f6aff2c2d57490751cb1a3b36daff173ecee74c809544d250fa9eafcb6906df60f540a9318bed9d1861daf44616af91f6c4a411ab28fa76138748ea7bc5eb03a815a85734272628ae2a323d528b45aa364254a5baa7267f7f8c9cdf8b4f1657c059981b5a59e4ffbbdfb6691eec501b241f70dfcb511a28c11a579b02944b9b87f537b84572dd8bee8632d632d21d815430d288aa604863288c1d2c461b07a7e16a020f173f4b28d93b16953eb0a0f7865f21f38e2463b3f7ae7af06774c817e66b1a116a0d27d72aed8a7b9da25e4ac817f97639f7c450376268a439c3a0ec158850b63e58e2873c305ca609e32c77cfa2c509e42b918c95d444405f593fa4cc722e96e7b9360d607c337598f96624d11126d00689602c6dee55359
  Actual:   61004061906a0040d00058006000506290690030d00059005d00b067e0640050cfc060cf55009063105f0010cfe059005604e0629065cc30c8004f08c00ac0cebce4c650cf6ce4009d0720a370a50280ce40930a6904f0751cacc2a0c2cca004220ba0cec0c90b20074027076c04f07080c4c380bb0018c57107a07560a00780c86046c82a0190bcf0f80190cddcefcd3608c03a80aa08a0c70cc7c7c10750ac10a9c88cc88cc4077a0a307a406b0590c89c8207860240a0ccf800a0c40cbcc7de13d0e240d6139011e0d212d91300dd70d812a011f0d212cc01eccefcce01ec019cc0015d02d02e1017029ccfec7400ec14c0eed0e714a013d0e713eb1520f340e114e01340e1133406c042e030ca90ceb02e01a00e3c940c9acfebc0eb9ad0bccfd0c31c6fce5ccf1c09cf49c81c6fab8ac81cb58b2ebbaf01b01a7cb8013cc81bf6c829c5fc2d6b95b4bbb70b45bcb7093062cc53087cc9cbe3c912cefcceeb6fcebcb8cb11be87cd2c4d4bfdbfebb68af3c868c37bb80b9dbe1bb54adfc287006087608f006000607f0085006089608f006000607f0080006092607d006000607d007ccff090f082cff0cff07ecf7b00ec83dc8300ac007c730071079c912c79071cc92c79c9c3028cc79cc4cdfcc96cb606220fbc3e1c2fcd4cc06c2503d30a007ce07a098c096cd509f805c10fb024c4bcbbdc73c634cf90ec7ccacf90cf7c6ccfeccf407d407ccf4ccf4c71cf930950972ce8093008ac08098d030082b07dc41cc42c8e02d803e01ffcedcf5cca4c0708140020149c7fce90c1bc43c56a0b106e10630b100b105d0b670ad06bb0670a9000b05f14c409d0c4d0c309d001d0bd11ee093cf25cef093c041ced0ead0aacaf8ca20a6c0a8ca20ab313acb75c9b12ec013c9b012010a032702a03c00830280b7e18ec804c0909accb5b110abecf50cff07bce9cceb011ceaec090d27cf7c090bd7c9bc3b3cd212140beccc0ccacefcd22c450176c88b31cb90c3abe500960fa8cea0880c98c72c961cf110f7cc3ceb0b85c63be80c9d0451cf5bc9cba5bedcc5bc3fcab0c90be9cb54bd2c20308500350080850085cff080208500450080850085cff080908800fc0010880064d000a510820592cae0820082caa08060e8006c0890e60062c8416fe0ea003000b0e0c0dacf60f2a029074f0ad0250cb8cea088d08a0b20cdf088cc2bbc9052a02002ae02ac750c3802acd1f095021b020c7c0be7020c9130710137013cb80be7013c433d00033ecf5c6c0beecf3cfdb04a038f07dcedcb85c96c68b00207ef07ac0dcc0dc8bcfb8c3ccb8f028c3acbc3c43ca15c970157c77c970bbbc39bff41070f800f90f900dc0f30f6507806fd06b06600510690604115104310511511130ff114804604e6020046004601e04a016f0a200a1163015e0a1169a14c0a890c6126012507c12bf09518bd13a089008d104081505f0fbb010c680b89c31c711082021f01ac730cc6014022cc9dc2c0c5bc75cb62bf9bcfc094d00ecff028cc63cf50273c98c83fbe7b8cbbadb93bfed0bbcf3bcec011cc33ce8ca28cc0c2c8c54cbccb68bf4bcdeca4cb25c5cbd8bb43b40c5bec07c14ebf7bb5bb2ab2bbe0a00110a110d0011002107000a00110a110d001100610a005a00811a010c0081d0010ed0030031074101cfe1cca0ff036400b0674061008000505b006400606cf05b0020cfc05d00e6052ce6dce604fccdfce60a54cbac5b8c5fcbacc2dc5904c0046cc68c86041cd00c8000f31080023c38cfbcc6ec3a015609c01e9c5309c0ca8bf3caba041cc0bc83041cc82c8008140740187c740710c5eb90c687058c873c85c6fcc6fc87057b022c7a6c71027cca9b86cad7048c93fc03030cc71bc1ca16053118b1080531cd41060b10cc5112110fcc51c1c10b06f306c116817906c006408806a1001164c151c370c4d0b4ce6d025072306802100140660137cfc03fb030cf00cda030cd6808b12cf0d808b00690e8094d05a034e039c70cb8dc4ac61313a020204e1410007ce40793051cbc5be004fcc27b83c8450e4048401f0e90cf0c480c79caec836be9caabc3eb99c95f10b0643cac1090020c3202b806bcb7dbc305dcc02b61c6d8cecca4cc57cf0bbedb4300bbcbfc0e8bf5c68bbd9b2fca0b08800b800f088008b007080b08800b800e088008a00b080a08b00e000d08b006b00b0b03087009f00908800510070b020be00d70040bc0048cff04fd194008800f050c008cfb016f081076fc910840083c90075e17906630360550c8ec0d0b22cf90247024c510c53020cf14c9c01c801bc9c0c48017cfddca309f7ca3c9e0c74c9fcd3bcc003fa003cc00bfecfd00ecc8fced004fc90cc24c74d0e3c88ce11cdfc94cc19ce100e1c41ce55058c3fcbcfc72cb83c9803f7caac9b0bc2c67c0300f613481340ec10ea1300f7606c17a00eb05e105a0e507251140246022114011101e11360bf037e009cfa0ce600a074414c1469140149114b13e144914e14d913c146114913b148c1390543cb606b001fcb405f50620e8a012c89cb93c13c79200509c804a006ccffce5cf22c94c498c68c95cb5fc09bcc001b02df05701cccaec950767c1ec6ecc12c1cbbb3bc1c09ecf90459021cf1ccb4cb702d8cd0c937c5dcc5bb65c04bcebcb0ca81c54be1bb41b4dc6c9c13c160bf9bbfbb29b38c00508700550050870089007080508700550050870089005080909b009902109b007ecdb07580880587ce60830051c790bfe020cfeccfd021c01ed000101cab0024006c9f0c8fcf2cab4c6b0b7d03dc6b0c5ec66c513bd90133cf0bdd0b47bb7c61c0410203020044000201c00940870a98070080cc68cca015e0a1020604e0a30ca3cf0ca1f04001d90a40430c7fc9808e506fcea8053071cc5cc6cc56ece30594055ce9cce7c70ce2508a025a0170460c7ac300582cfe03ebca8cff0c2cc6ac60406d0120015065005600106f906700a200f05dc052cf4066a06206c906b065005c06706b207bcaecc83070c04cc8504fd06a008ccff063c05dcfd05f8043015acfd02dccfecf904ccc930cca032c840c86032c864bea05f3c89be2ca88b84b8450ad04930780b50cb500fcb330c1056607309c0c89015ce0f0af00e907a03b0c81ca705930d3090c038cd4cceace803290fa029002d04d0c8902d00cb12608bd057067cc71cf9cc8b045052fcffc82cbf4befcf66cafcb87c9ac61cbd6bd6c93cbc8a34a6130d6d2ad691bd6051c643bfd5df6308f1f0777b432442a22ba84771d935de99bb7b27da103b10817a4a5e8d135f53b5064d21a432f24be0e21e98a9eeb13c934b00880afb3c10682afdee4afa016e63a3e858e3a1a89ce257aef90a00110a110d0011002107000a00110a110d001100610a005a00811a010c0081d0010ed0030031074101cfe1cca0ff036400b0674061008000505b006400606cf05b0020cfc05d00e6052ce6dce604fccdfce60a54cbac5b8c5fcbacc2dc5904c0046cc68c86041cd00c8000f31080023c38cfbcc6ec3a015609c01e9c5309c0ca8bf3caba041cc0bc83041cc82c8008140740187c740710c5eb90c687058c873c85c6fcc6fc87057b022c7a6c71027cca9b86cad7048c93fc03030cc71bc1ca16053118b1080531cd41060b10cc5112110fcc51c1c10b06f306c116817906c006408806a1001164c151c370c4d0b4ce6d025072306802100140660137cfc03fb030cf00cda030cd6808b12cf0d808b00690e8094d05a034e039c70cb8dc4ac61313a020204e1410007ce40793051cbc5be004fcc27b83c8450e4048401f0e90cf0c480c79caec836be9caabc3eb99c95f10b0643cac1090020c3202b806bcb7dbc305dcc02b61c6d8cecca4cc57cf0bbedb4300bbcbfc0e8bf5c68bbd9b2fca0b08800b800f088008b007080b08800b800e088008a00b080a08b00e000d08b006b00b0b03087009f00908800510070b020be00d70040bc0048cff04fd194008800f050c008cfb016f081076fc910840083c90075e17906630360550c8ec0d0b22cf90247024c510c53020cf14c9c01c801bc9c0c48017cfddca309f7ca3c9e0c74c9fcd3bcc003fa003cc00bfecfd00ecc8fced004fc90cc24c74d0e3c88ce11cdfc94cc19ce100e1c41ce55058c3fcbcfc72cb83c9803f7caac9b0bc2c67c0300f613481340ec10ea1300f7606c17a00eb05e105a0e507251140246022114011101e11360bf037e009cfa0ce600a074414c1469140149114b13e144914e14d913c146114913b148c1390543cb606b001fcb405f50620e8a012c89cb93c13c79200509c804a006ccffce5cf22c94c498c68c95cb5fc09bcc001b02df05701cccaec950767c1ec6ecc12c1cbbb3bc1c09ecf90459021cf1ccb4cb702d8cd0c937c5dcc5bb65c04bcebcb0ca81c54be1bb41b4dc6c9c13c160bf9bbfbb29b38c00508700550050870089007080508700550050870089005080909b009902109b007ecdb07580880587ce60830051c790bfe020cfeccfd021c01ed000101cab0024006c9f0c8fcf2cab4c6b0b7d03dc6b0c5ec66c513bd90133cf0bdd0b47bb7c61c0410203020044000201c00940870a98070080cc68cca015e0a1020604e0a30ca3cf0ca1f04001d90a40430c7fc9808e506fcea8053071cc5cc6cc56ece30594055ce9cce7c70ce2508a025a0170460c7ac300582cfe03ebca8cff0c2cc6ac60406d0120015065005600106f906700a200f05dc052cf4066a06206c906b065005c06706b207bcaecc83070c04cc8504fd06a008ccff063c05dcfd05f8043015acfd02dccfecf904ccc930cca032c840c86032c864bea05f3c89be2ca88b84b8450ad04930780b50cb500fcb330c1056607309c0c89015ce0f0af00e907a03b0c81ca705930d3090c038cd4cceace803290fa029002d04d0c8902d00cb12608bd057067cc71cf9cc8b045052fcffc82cbf4befcf66cafcb87c9ac61cbd6bd6c93cbc8a34a6130d6d2ad691bd6051c643bfd5df6308f1f0777b432442a22ba847
 ```

Anyways, I think I will consider progress blocked until we can get some clarity from the PQC team on the following two questions:

1. Which tests are valid/maintained?
2. Is the methodology of a static .s file + .exp file still adhered to?
 -- If so, where are the .exp files?
 -- If not, maybe we should align on the goals of the CI tests. While on the one hand I can see value in having a fresh bit of randomness injected in every test while vetting an algorithm, my impression is that a nightly-style CI would want fixed, static numbers (even if they are supposed to be random), because the idea is to sacrifice coverage in favor of reproducibility in a nightly style build. However, I don't actually fully understand what it means to test a _post-quantum_ algorithm, so maybe some amount of unreproducibility is inherent in testing anything quantum?

**Comment by jadephilipoom** [comment]
> @jadephilipoom would a reasonable next step be to figure out how to get these test running using that bazelisk command?

Looks like one of the `#define`s didn't fire, which looks like missing `--copt`? Let me try locally.

**Comment by jadephilipoom** [comment]
Huh, weirdly I'm actually getting toolchain issues on this branch. 
```
Unable to find 'riscv32-unknown-elf-gcc' in PATH or in '/tools/riscv/bin'. Set the 'RV32_TOOL_GCC' or TOOLCHAIN_PATH environment variable if you installed your RISC-V toolchain in an alternate location.
Target //sw/device/tests/crypto:aes_functest_fpga_cw310_rom_with_fake_keys failed to build
```
The same target builds with no problem on `expo/otbn-pqc` and is unrelated to these changes. I don't see anything in the diff that looks especially suspicious, but maybe some of the new code relies on the old RISC-V toolchain setup (which was configured through environment variables rather than Bazel). Mostly logging here in case anyone else hits something similar later. Debugging now.

PR #11, opened 2025-06-05
[otbn] Update tooling to support simulating existing PQC code.

jadephilipoom:
This is the same content as https://github.com/zerorisc/expo/pull/8 but with a linear commit history that passes the OpenTitan commit linter (and adding https://github.com/bunnie/expo/pull/9 on top).

CC @bunnie


bunnie:
poked around and this looks great to me! thanks for putting this all together, sorry I kind of dropped off for the last mile.

PR #65, opened 2025-09-03
[topgen] Removing alert and cpu crashdumps and CSRs from rstmgr

LouisTheLuis:
Removing the `alert_handler` and/or the `rv_core_ibex` from the top config should remove the corresponding crashdumps and CSRs.


hw/ip_templates/rstmgr/data/rstmgr.hjson.tpl
```diff
@@ -179,9 +182,11 @@
     { name: "RSTMGR.CPU_INFO.ENABLE",
       desc: "Enable capture of cpu crash dump information."
     }
+% if with_alert_handler:
```
matutem, line 185: I think this is orthogonal to the lack of alert_dump. It is possible there is a rstmgr that is not the ROT's (say it is from subsystem XYZ), and the alert dump is sent to the ROT's rstmgr. But there could be alert senders in XYZ, which means they would need to respond to alert sender pings, so the ROT's alert handler would need to know then those senders are unavailable, thus XYZ's rstmgr would need to send RESET_STATUS for those alert senders.

hw/ip_templates/rstmgr/data/rstmgr.hjson.tpl
```diff
@@ -167,9 +168,11 @@
     { name: "RSTMGR.RESET_INFO.CLEAR",
       desc: "Clear information about the causes of a reset."
     }
+% if with_alert_dump:
     { name: "RSTMGR.ALERT_INFO.CAPTURE",
       desc: "Capture alert crash dump information upon reset."
     }
+% endif
```
matutem, line 175: I think this endif should move to line 179, since both alert info capture and enable only apply if `with_alert_dump`.

hw/ip_templates/rstmgr/data/rstmgr.hjson.tpl
```diff
@@ -167,9 +168,11 @@
...
     { name: "RSTMGR.ALERT_INFO.ENABLE",
       desc: "Enable capture of alert crash dump information."
     }
```
matutem, line 178: Please also surround cpu_info capture and enable in it `with_cpu_dump`.

hw/ip_templates/rstmgr/data/rstmgr.tpldesc.hjson
```diff
@@ -124,11 +125,17 @@
       default: {}
     }
     {
-      name: "with_alert_handler"
+      name: "with_alert_dump"
       desc: "Generate outputs for a rstmgr that would connect to an alert handler"
```
matutem, line 129:
```suggestion
      desc: "Generate inputs and CSRs to capture an alert dump."
```

hw/ip_templates/rstmgr/doc/theory_of_operation.md.tpl
```diff
@@ -134,7 +134,7 @@ The reset manager then checks as follows:
 
 - If all reset conditions are satisfied, wait for the reset release to gracefully complete the cycle.
 
-% if with_alert_handler:
+% if with_alert_dump:
```
matutem, line 137: We discussed the reset indications are independent of handling an alert dump, so line 185 in rstmgr.hjson.tpl is not conditioned by `with_alert_dump`. On the same vein, this section should not be conditioned.

hw/ip_templates/rstmgr/rstmgr.core.tpl
```diff
@@ -8,7 +8,7 @@ description: "Reset manager RTL"
 filesets:
   files_rtl:
     depend:
-% if with_alert_handler:
+% if with_alert_dump:
       - ${instance_vlnv("lowrisc:ip:alert_handler_pkg")}
 % endif
       - lowrisc:ip:rv_core_ibex_pkg
```
matutem, line 14: I imagine `rv_core_ibex_pkg` is only needed for the cpu dump, so this dependency should be conditioned by `with_cpu_dump`.

hw/ip_templates/rstmgr/data/rstmgr.hjson.tpl
LouisTheLuis: Included condition for `with_cpu_dump`. Done.

hw/ip_templates/rstmgr/data/rstmgr.hjson.tpl
LouisTheLuis: Understood.

hw/ip_templates/rstmgr/data/rstmgr.tpldesc.hjson
LouisTheLuis: Done.

hw/ip_templates/rstmgr/rstmgr.core.tpl
LouisTheLuis: Done.

PR #66, opened 2025-09-03
[RFC, DV] Reset Management in Verification Testbenches

ysekhar:
This RFC outlines standard practices, do's, and dont's for effective reset management in verification testbenches employing Universal Verification Methodology in SV-UVM or pyUVM testbenches. 

Proper reset handling is crucial for reliable testbench operation, accurate DUT (Design Under Test) initialization, and efficient debug.


rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+RFC aims to standardize reset handling to improve testbench robustness, reusability, and 
+debuggability.
+
+## Problem Statement
+Testbenches often involve intricate reset scenarios, such as power-on resets, software-initiated 
+resets, or partial resets in multi-domain systems. Inconsistent handling of resets can lead to 
+simulation artifacts, race conditions, or unverifiable states.
+
+Common issues with current reset management approach include:
+* Lack of standardized reset coordination mechanisms
```
qmn, line 29: I think some examples are good, just to make these more concrete. I believe one thing to say here is the use of [stress_all_with_rand_reset](https://github.com/zerorisc/expo/blob/ccefb891dfc8a0e6a39fddde06798fd9dba46388/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L792), which bolts a random reset thread onto an existing sequence. Then there's [apply_resets_concurrently](https://github.com/zerorisc/expo/blob/ccefb891dfc8a0e6a39fddde06798fd9dba46388/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L162), which seems to need to know something about EDNs in the base sequence...

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+* Race conditions between reset deassertion and test stimulus
```
qmn, line 30: Among other things, this can arise when you use hardcoded delays between reset de-assert and some other check / action?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+* Inconsistent reset handling across different verification components
+
+# Goals
```
qmn, line 33: I think it's important to say why these goals matter. To me, aside from the benefits of having good reset handling, it's about:

1. making reset testing easy, so you don't avoid it;
2. making reset sequencing a first-class consideration from the beginning;
3. improves testing posture because you can (soon) reason about complex interactions between different reset domains
4. simplifies debug because you can easily disentangle reset from other effects.

What do you think?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+This RFC outlines best practices for reset assertion, de-assertion, sequencing, and synchronization 
+within the UVM framework.:
+* A unified reset management framework for UVM testbenches 
+* Standard interface and protocol for reset implementation / co-ordination
+* Potentially reusable clk & reset component
+
+# Standard Practices for Reset Management
```
qmn, line 40: I found this section hard to comprehend -- is it to say that you are prescribing features of a competent reset management scheme? In other words, how do I read this section?

It's worth telegraphing at this point in the RFC that it's from these standard practices that you construct your list of Do's and Don'ts, and it's what guides your implementation in #72, #74.

I think two additional diagrams focusing on reset-aware components would be really meaningful: one that depicts functional interactions at a single block level, and perhaps one showing hierarchical propagation. This spatial / functional organization would complement the current reset_flow diagram, which shows a more temporal organization.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+**Single Source of Truth:** A testbench should have a single, well-defined point for controlling 
+and asserting/deasserting the DUT's reset signal(s). Typically handled within the base_test or a 
+dedicated base_seq for the block.
+
+**Hierarchical Propagation:** Reset should propagate downwards through the testbench/environment 
+hierarchy (e.g. from sequence to reset agent to reset driver). All testbench elements should then 
+synchronize with either reset interface (drivers/monitors - at a pin level) or the reset monitor 
+(all other components - at a transaction level) to ascertain the state of reset in the testbench and 
+respond when reset is triggered.
```
qmn, line 52: > respond

Who, the driver/monitor at pin level or other components at the transaction level?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Synchronous vs. Asynchronous Resets
+
+* Assertion of reset (going active) should occur immediately, without waiting for the clock edge.  
+* De-assertion (release) synchronized to the rising edge of the clock to prevent metastability.
```
qmn, line 57: Just curious, are these practices already observed, or will we need to implement this?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+**Understand DUT Behavior:** Clearly understand whether the DUT's reset is synchronous or 
+asynchronous, and whether it's active-high or active-low. Pick one polarity and stick to it 
+consistently everywhere, e.g. use reset\_n (active-low) for DUT and TB.
```
qmn, line 61: The [style guide](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#resets) prescribes active-low and `rst_n` for the name, which I think is fine. We can point to the style guide?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+The testbench reset mechanism must accurately mimic this behavior. 
+
+**Clock-Gated Resets:** If the DUT uses clock-gated resets, ensure the testbench models this
```
qmn, line 65: > clock-gated resets

Just curious, anyone know if these exist in the repo today?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+The testbench reset mechanism must accurately mimic this behavior. 
+
+**Clock-Gated Resets:** If the DUT uses clock-gated resets, ensure the testbench models this 
+correctly, accounting for potential setup/hold violations during reset deassertion.
+
+**Global vs. Local resets:** 
+
+* Global reset: whole-chip/system reset (e.g. power-on).  
+* Local resets: block/module-level resets. Use only if DUT has isolated reset domains.
```
qmn, line 71: Not sure what this section is about, is it to state something about how agents should be organized?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+transactions and return to a clean state.
+
+**Monitors:** Monitors should also be reset-aware (at a pin level). When a monitor see the assertion
+of reset signal it clear out any active data collection and return to a clean state. It should  then 
+re-synchronize with the DUT after reset deassertion to restart interface sampling and transaction
+sampling.
+
+**Synchronization & Metastability:**  Monitors need logic similar to dual-flop synchronizer on any 
+asynchronous reset input crossing clock domains. Avoid direct sampling of async de-assert without 
+sync. Clearly document which resets are async vs sync.
```
qmn, line 86: This one doesn't seem to be a component per se. By the way, what's the current use / testing for metastability?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+**Scoreboards/Checkers:** Scoreboards and checkers should clear their expected and actual data 
+models during reset to avoid false mismatches. Preferred method of connecting the scoreboard to the 
+reset monitor should be via callbacks, however analysis ports can also be used. (The recommendations 
+are specific to UVM methodology)
```
qmn, line 91: > Preferred method of connecting the scoreboard to the reset monitor should be via callbacks, however analysis ports can also be used. (The recommendations are specific to UVM methodology)

This seems to be an implementation detail, but it raises an important point: why _should_ we use callbacks over analysis ports?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+**Sequences:**  Resets should only be driven from the master sequence running on the virtual 
+sequencer. When in reset no secondary sequences should be active or attempt to generate any 
+transactions. The relevant secondary sequence should only start execution after the DUT is out of 
+reset and stable. During normal test execution, when a reset is triggered randomly, the virtual 
+sequence should wait unitl it is clear that reset is issued to the DUT and then terminate all 
+secondry sequences and go back to restart execution only when the DUT is out of reset and stable.
```
qmn, line 98: The second half of this paragraph seems to say the same as the first half, unless I'm missing a nuance

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+**Virtual Sequences:** Virtual sequences should coordinate reset application across multiple agents 
+or if the DUT has multiple reset domains.
+
+**Coverage & Checks:**  
+
+* Cover “reset asserted” event.  
+* Cover “reset de-assert coincident with X line” if DUT has timing dependencies.
```
qmn, line 106: These also don't seem to be components per se, but no worries.

> Cover "reset asserted" event.

Would you say one for each domain?

> Cover "reset de-assert coincident with X line"

Is "X line" an arbitrary net?

Also, nit, but these look like smart quotes / not ASCII.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+Self-checking: after de-assert, verify that all DUT registers have their default values (Can be
```
qmn, line 108: Just wondering, is there a suggestion to change how self-checking is done?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+**Coverage & Checks:**  
+
+* Cover “reset asserted” event.  
+* Cover “reset de-assert coincident with X line” if DUT has timing dependencies.
+
+Self-checking: after de-assert, verify that all DUT registers have their default values (Can be 
+randomized such that not every test needs to bear the overhead)
+
+Exceptions: OTP, Flash, and ReRAMs need to prove they are nonvolatile and may have other 
+functionality to verify too at the point of reset de-assertion.
```
qmn, line 112: > other functionality

Any chance you could give more flavor?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Reset Sequence
+
+**Common Reset Sequence:** Should be a generic sequence that all TBs can use to drive reset. It is 
+transaction driven i.e. A reset transaction is created in the reset sequence and sent down via the 
+sequencer stack like any other transaction.
```
qmn, line 118: A+. Crucially, not just a separate sequence that is invoked only ~~~sometimes~~~.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+
+It should have the following control knobs: 
+* Assert-time duration  
+* Randomized delays post de-assert: small random offset after reset to expose corner-cases.  
+* Clock alignment: always align reset de-assert to clock edge.
```
qmn, line 123: This seems to presume the current system, which is to async assert and sync de-assert?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+**Test-Specific Resets:** Allow for execution of test-specific reset sequences if necessary (e.g.
+asserting reset mid-test for specific error injection scenarios), but these should build upon the 
+standard reset mechanism.
+
+**Execution of Reset Sequence:** Incorporate a dedicated reset sequence invocation at the beginning 
+of every test. This should:
+* Assert reset for a specified duration.  
+* Deassert reset.  
+* Wait for a sufficient (random) number of clock cycles to allow the DUT and testbench components to 
+  stabilize post-reset.
```
qmn, line 139: > dedicated reset sequence invocation

Crucially, this would be very similar to a randomly triggered mid-test reset.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+    * use a virtual interface (vif\_proxy) for the reset signal to provide a clean, abstract way for 
+      testbench components to monitor the reset.  
+* Parameterize/reset timing (pulse width, delay) so different tests can override.  
+* Synchronize de-assertion to clock — never use zero-delay (\# delays) assignment.  
+* Randomize reset skew (within DUT spec) to catch corner cases.  
+* Include post-reset functional checks in scoreboards   
+  * These are usually triggered by sequences doing access to CSRs  
+  * Explicit CSR tests need to readback all registers just after reset  
+* When reset is observed during operational testcases, readback of registers is not always required, 
+  but tests have to restart. If changes were observed to Flash / ReRAMs verify the state is correct.
```
qmn, line 171: I think the big picture is that non-volatile elements (and really, anything that responds differently to a reset) needs to be checked, if I understand right?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+* Synchronize de-assertion to clock — never use zero-delay (\# delays) assignment.  
+* Randomize reset skew (within DUT spec) to catch corner cases.  
+* Include post-reset functional checks in scoreboards   
+  * These are usually triggered by sequences doing access to CSRs  
+  * Explicit CSR tests need to readback all registers just after reset  
+* When reset is observed during operational testcases, readback of registers is not always required, 
+  but tests have to restart. If changes were observed to Flash / ReRAMs verify the state is correct.   
+* Collect reset coverage in block/top level functional coverage plan. 
+
+# Don’ts
```
qmn, line 174: I found myself confused in this section, roughly at "Don't forget to disable or clean up". Is that "[Don't] don't forget..."? May I suggest each line get a "don't" in front?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+  active stimulus.  
+* Neglect the impact of reset on verification IP (VIP) or third-party models. Ensure they are also 
+  properly reset and synchronized.  
+* Forget to verify what happens if reset is asserted during an ongoing transaction or in unusual 
+  corner cases.  
+* Assume the DUT is in a clean state just because reset was asserted. Always perform post-reset 
+  checks (There are overheads here that need to be considered)
+
+# Future Topics / Explorations
+* Multi-Domain Support
```
qmn, line 199: I think multi reset domain support is going to need to come sooner.

The other big thing is that alerts and interrupts also have broad state-altering behavior not unlike a reset. (This idea due to Matute.) I wonder if the same technique to separate resets this way can also be applied to alerts and interrupts.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
```diff
@@ -0,0 +1,201 @@
...
+re-synchronize with the DUT after reset deassertion to restart interface sampling and transaction
+sampling.
+
+**Synchronization & Metastability:**  Monitors need logic similar to dual-flop synchronizer on any 
+asynchronous reset input crossing clock domains. Avoid direct sampling of async de-assert without 
+sync. Clearly document which resets are async vs sync.
+
+**Scoreboards/Checkers:** Scoreboards and checkers should clear their expected and actual data 
+models during reset to avoid false mismatches. Preferred method of connecting the scoreboard to the 
+reset monitor should be via callbacks, however analysis ports can also be used. (The recommendations
```
apinise, line 90: > Preferred method of connecting the scoreboard to the reset monitor should be via callbacks, however analysis ports can also be used.

I understand how the scoreboard to reset monitor connection would work with analysis ports, but am not so clear with the callback methodology. What are the advantages / reasons for callbacks being preferred?

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: I believe the above two are in play in OT.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: OT already has various reset domains. Agent instance need to be organized accordingly

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: Yes

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: Usually  done via assertions

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: Typically there is a test case that does this explicitly. Not all tests need to prove register default values. Doing so will create a lot of redundant cycles that are not needed. 

Not suggesting a change to the current flow but adding a checklist point to be ticked off.

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: Yes. But the TB should be able to drive the de-assertion asynchronously. Its the design that needs to make sure internal reset release is sync to a clock edge

rfc/rfc-2025-01-DV-Reset-Management/rfc_2025_01_dv_reset_management.md
ysekhar: Ok what confused you. The section is marked as "Don't" .

qmn (approved):
Likewise, I support this RFC. Thanks, Yogish!

PR #71, opened 2025-09-17
[bazel] Make latest bitstream default as expo-specific Bazel fix

pqcfox:
This PR is a one-line fix which makes use of the latest bitstream default to prevent issues with builds in expo.


RyanTorok (approved):
This is fine for now. The one thing to keep in mind is that is upstream makes changes that we haven't adopted yet in a weekly sync, we run the risk of the bitstream getting _ahead_ of our branch.

PR #91, opened 2025-10-09
[crypto] Add missing RSA PSS maskedDB MSB check

pqcfox:
This tiny PR adds the missing maskedDB MSB check required by [RFC 8017 section 9.1.2](https://www.rfc-editor.org/rfc/rfc8017#section-9.1.2) step 6.

```
 6.  If the leftmost 8emLen - emBits bits of the leftmost octet in
     maskedDB are not all equal to zero, output "inconsistent" and
     stop.
```


pqcfox:
Closed as redundant with #95.

PR #103, opened 2025-10-22
[util,ipgen] Improve file handling

matutem:
Catch errors when config file cannot be read or render fails.

### DIFF
--- a/util/ipgen.py
+++ b/util/ipgen.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 r"""IP Generator: Produce IP blocks from IP templates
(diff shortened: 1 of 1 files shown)


util/ipgen.py
```diff
@@ -28,15 +29,23 @@ def action_generate(ip_template: IpTemplate, args: argparse.Namespace) -> None:
...
     # Read the IP configuration file.
-    config_fp = args.config_file
-    config_text = config_fp.read()
-    config_fp.close()
+    try:
+        config_text = args.config_file.read_text(encoding="utf-8")
+    except Exception as e:
+        logging.error(f"Could not read config file: {e}")
+        sys.exit(1)
+
```
qmn: Unlike below, this Exception (perhaps IOError) isn't caught by the try...catch in `__main__`. We could either 1) hoist the try...catches into here and do away with the one in `__main__` entirely, or 2) expand the list of exceptions caught in `__main__`

util/ipgen.py
```diff
@@ -28,15 +29,27 @@ def action_generate(ip_template: IpTemplate, args: argparse.Namespace) -> None:
...
     # Render the IP template into an IP block.
-    renderer = IpBlockRenderer(ip_template, ip_config)
-    renderer.render(output_path, overwrite_output_dir)
+    try:
+        renderer = IpBlockRenderer(ip_template, ip_config)
+        renderer.render(output_path, overwrite_output_dir)
+    except TemplateRenderError as e:
+        logging.error(e.verbose_str())
+        sys.exit(1)
+
```
qmn: flake8 says "E303 too many blank lines"

```suggestion
```

PR #107, opened 2025-10-30
[crypto] Fix a broken test.

jadephilipoom:
The test configuration we inherited from upstream loads a couple of totally unused variables, which just gets ignored for upstream because they load the maximal p384 memory layout for all tests. Since we only include symbols in the binary for the buffers that are actually used, though, the symbols for d0/d1 don't appear and cause an error. The commit 3435eb introduced this break.

@rtorok-zr noticed this one during CI setup, CI working as intended!


pqcfox (approved):
LGTM, thanks!

PR #109, opened 2025-10-31
[crypto] Mini sync of P-384 on-curve check fault fix

pqcfox:
This PR does a one-off sync of upstream commit `c690c4a037d2bfb0ae80fdd7da62193dcfde1fea` to repair the broken ECDH KAT test, ensuring that off-curve public keys don't cause a fault but simply yield an invalid operation.


rtorok-zr (approved):
Thanks Kat!

PR #126, opened 2025-11-14
[crypto] Add instruction count runtime checks based on static analysis

pqcfox:
This PR adds runtime instruction count checks for OTBN programs called from cryptolib. The bounds for valid instruction counts are computed in a build-time static analysis step, and these bounds are inserted into templated header files which can be accessed by the cryptolib implementation source files in `sw/device/lib/crypto/impl`.

In order to support OTBN programs with multiple "modes," e.g. different RSA key sizes in `run_rsa_modexp.s` selected using the dmem `mode` buffer, a HJSON file is input to the static analysis step indicating which imem symbol corresponds to which program mode (i.e. `rsa_2048_modexp` being the program symbol for generated constants `kRsa2048Modexp{Min,Max}InstructionCount`). The static analysis script ensures that only control flow paths which flow through the PC corresponding to this symbol are counted when computing the instruction count range.

As such, this PR includes

- various modifications to the existing instruction count range algorithm to achieve sharper instruction bounds
- a new script for auto-generating headers based on instruction counts from this algorithm
- a new Bazel build rule, `autogen_otbn_insn_count_header`, for building headers with instruction count constants
- a new macro in `crypto/drivers/otbn.h`, `OTBN_CHECK_INSN_COUNT`, for hardened instruction count checking
- modifications to the implementations in `/sw/device/lib/crypto/impl` to make use of this macro

Together, this means that with this PR, P-256, P-384, Ed25519, SHA-256, and SHA-512 now have all operations paired with an automatically-generated instruction count check. RSA has all operations based on the modexp routine checked as well; RSA keygen requires further alterations to the static analysis pass to handle one small part of the key generation logic.

### DIFF
--- a/hw/ip/otbn/util/shared/information_flow_analysis.py
+++ b/hw/ip/otbn/util/shared/information_flow_analysis.py
@@ -274,8 +274,6 @@ def _get_iflow_update_state(
         else:
             loop_iters[pc] = rec_loop_iters[pc]
 
-    loop_iters.update(rec_loop_iters)
-
     # Update the constants to keep only those that are either unmodified in the
     # return-path information flow or returned by the recursive call.
     constants.removemany(rec_return_iflow.all_sinks())
--- a/sw/device/lib/crypto/impl/rsa/rsa_signature.c
+++ b/sw/device/lib/crypto/impl/rsa/rsa_signature.c
@@ -169,7 +169,7 @@ status_t rsa_signature_generate_2048_start(
 }
 
 status_t rsa_signature_generate_2048_finalize(rsa_2048_int_t *signature) {
-  return rsa_modexp_consttime_2048_finalize(signature);
+  return rsa_modexp_consttime_crt_2048_finalize(signature);
 }
 
 /**
@@ -301,7 +301,7 @@ status_t rsa_signature_generate_3072_start(
 }
 
 status_t rsa_signature_generate_3072_finalize(rsa_3072_int_t *signature) {
-  return rsa_modexp_consttime_3072_finalize(signature);
+  return rsa_modexp_consttime_crt_3072_finalize(signature);
 }
 
 /**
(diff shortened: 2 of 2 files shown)


sw/device/lib/crypto/impl/ecc/p256.h
```diff
@@ -278,6 +279,21 @@ status_t p256_ecdh_finalize(p256_ecdh_shared_key_t *shared_key);
...
 status_t p256_sideload_ecdh_start(const p256_point_t *public_key);
 
+/**
+ * Finish an async ECDH/P-256 shared key generation operation on OTBN.
+ *
+ * Uses a private key generated from a key manager seed. The key manager should
+ * already have sideloaded the key into OTBN before this operation is called.
+ *
+ * Blocks until OTBN is idle. May be used after either `p256_ecdh_start` or
+ * `p256_sideload_ecdh_start`; the operation is the same.
```
jadephilipoom: The instruction count checks are now different between these two; the second sentence here should probably be removed both in this doc comment and in the one for `p256_ecdh_finalize`.

sw/device/lib/crypto/impl/ecc/p384.h
```diff
@@ -278,6 +279,21 @@ status_t p384_ecdh_finalize(p384_ecdh_shared_key_t *shared_key);
...
 status_t p384_sideload_ecdh_start(const p384_point_t *public_key);
 
+/**
+ * Finish an async ECDH/P-384 shared key generation operation on OTBN.
+ *
+ * Uses a private key generated from a key manager seed. The key manager should
+ * already have sideloaded the key into OTBN before this operation is called.
+ *
+ * Blocks until OTBN is idle. May be used after either `p384_ecdh_start` or
+ * `p384_sideload_ecdh_start`; the operation is the same.
```
jadephilipoom: Same comment here as for P256.

sw/otbn/crypto/montmul.s
```diff
@@ -575,6 +575,12 @@ mont_loop:
  * clobbered Flag Groups: FG0, FG1
  */
 montmul:
+  /* set pointers to allow static analysis constant propagation */
+  /* TODO (#137): find a way to elide this while allowing instruciton count
```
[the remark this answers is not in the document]
pqcfox: Fixed, thanks!

pqcfox:
Rebased on #154 to get initial CI lint check working, do not merge until #154 is merged.

PR #159, opened 2025-11-24
Modify cryptolib `*_async_finalize` operations to return `OTCRTYPO_ASYNC_INCOMPLETE`

pqcfox:
This PR modifies the `*_async_finalize` functions for OTBN operations in cryptolib to return `OTCRYPTO_ASYNC_INCOMPLETE` instead of busy waiting.

The synchronous counterparts to `*_async_finalize` which used the async API internally were correspondingly modified to force a busy wait before calling the proper finalize function.

### DIFF
--- a/sw/device/tests/crypto/ecdsa_p256_verify_functest.c
+++ b/sw/device/tests/crypto/ecdsa_p256_verify_functest.c
@@ -34,6 +34,7 @@ status_t ecdsa_p256_verify_test(
   TRY(p256_ecdsa_verify_start(&testvec->signature, digest.data,
                               &testvec->public_key));
   hardened_bool_t result;
+  TRY(otbn_busy_wait_for_done());
   TRY(p256_ecdsa_verify_finalize(&testvec->signature, &result));
 
   if (testvec->valid && result != kHardenedBoolTrue) {


sw/device/lib/crypto/impl/rsa/rsa_signature.c
```diff
@@ -222,7 +223,13 @@ status_t rsa_signature_verify_2048_finalize(
     hardened_bool_t *verification_result) {
   // Wait for OTBN to complete and get the size for the last RSA operation.
```
jadephilipoom: nit: this comment should probably change in all 3 places in this file to indicate that it no longer waits for OTBN, but rather fails if OTBN isn't done.

jadephilipoom (approved):
LGTM, thanks Kat!

pqcfox:
Running `cryptolib_tests.sh` locally, so far looks like `//sw/device/tests/crypto:ecdsa_p256_verify_functest_hardcoded_fpga_cw310_rom_with_fake_keys` and a few other tests may need some updating. Will let the run finish and then will add fixes (likely w/ use of the async APIs) to the tests.

PR #168, opened 2025-12-10
Add RSA key import and export APIs

pqcfox:
This PR adds RSA key import and export APIs beyond the ones that were already present. Namely, this PR adds API interfaces for:

- RSA public key export
- RSA private key import with validity checks
- RSA private key export

Follow-up tasks for next PRs (as this one is already plenty large):

- Update `otcrypto_interface.{c,h}` to include functions added in this PR
- Tweak the instruction count tracking logic to also autogen headers + do runtime checks for private key check OTBN calls
- Convert RSA private key import and check logic to return `OTCRYPTO_ASYNC_INCOMPLETE` instead of spinning

### DIFF
--- a/sw/device/lib/crypto/impl/rsa.c
+++ b/sw/device/lib/crypto/impl/rsa.c
@@ -245,6 +245,8 @@ otcrypto_status_t otcrypto_rsa_public_key_deconstruct(
   HARDENED_TRY(public_key_structural_check(public_key));
 
   // Extract the values of n and e based on the inferred size.
+  // Note that we use `hardened_memcpy` as adefense against fault injection, as
+  // faulting RSA public keys to weak moduli is often quite easy.
   switch (size) {
     case kOtcryptoRsaSize2048: {
       if (modulus.len != kRsa2048NumWords) {
@@ -615,6 +617,10 @@ otcrypto_status_t otcrypto_rsa_private_key_deconstruct(
       return OTCRYPTO_FATAL_ERR;
     }
   }
+
+  // Should be unreachable.
+  HARDENED_TRAP();
+  return OTCRYPTO_FATAL_ERR;
 }
 
 otcrypto_status_t otcrypto_rsa_keypair_from_cofactor(
--- a/sw/device/lib/crypto/include/rsa.h
+++ b/sw/device/lib/crypto/include/rsa.h
@@ -226,7 +226,7 @@ otcrypto_status_t otcrypto_rsa_private_key_construct_and_check_async_finalize(
     otcrypto_blinded_key_t *private_key, hardened_bool_t *key_valid);
 
 /**
- * Deonstructs an RSA private key into cofactors of the modulus, private
+ * Deconstructs an RSA private key into cofactors of the modulus, private
  * exponent CRT components, and a CRT coefficient.
  *
  * The caller should allocate space for the private key and set the `keyblob`,
@@ -238,7 +238,7 @@ otcrypto_status_t otcrypto_rsa_private_key_construct_and_check_async_finalize(
  * @param[out] d_p First CRT component of the RSA private exponent d (d_p).
  * @param[out] d_q Second CRT component of the RSA private exponent d (d_q).
  * @param[out] i_q CRT reconstruction coefficient for given cofactors (i_q).
- * @return Result of the RSA key construction.
+ * @return Result of the RSA key deconstruction.
  */
 otcrypto_status_t otcrypto_rsa_private_key_deconstruct(
     otcrypto_blinded_key_t *private_key, otcrypto_word32_buf_t p,
--- a/sw/otbn/crypto/rsa_keygen.s
+++ b/sw/otbn/crypto/rsa_keygen.s
@@ -179,13 +179,13 @@ rsa_keygen:
  * @param[in]  w31: all-zero
  * @param[out] dmem[rsa_n..rsa_n+(plen*2*32)] RSA public key modulus (n)
        check value
- * @param[in]  dmem[rsa_d_p..rsa_d_p+(plen*32)] first RSA private key private
+ * @param[out]  dmem[rsa_d_p..rsa_d_p+(plen*32)] first RSA private key private
        exponent CRT component (d_p) check value
- * @param[in]  dmem[rsa_d_q..rsa_d_q+(plen*32)] second RSA private key private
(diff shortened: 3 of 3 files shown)


sw/device/lib/crypto/include/rsa.h
```diff
@@ -120,28 +120,130 @@ otcrypto_status_t otcrypto_rsa_public_key_construct(
...
+ * The caller should allocate space for the private key and set the `keyblob`,
+ * `keyblob_length`, and `key_length` fields accordingly.
+ *
+ * @param private_key Source private key struct.
+ * @param[out] p First cofactor of RSA modulus (p).
+ * @param[out] q Second cofactor of RSA modulus (q).
+ * @param[out] d_p First CRT component of the RSA private exponent d (d_p).
+ * @param[out] d_q Second CRT component of the RSA private exponent d (d_q).
+ * @param[out] i_q CRT reconstruction coefficient for given cofactors (i_q).
+ * @return Result of the RSA key construction.
```
jadephilipoom:
```suggestion
 * @return Result of the RSA key deconstruction.
```

sw/device/lib/crypto/impl/rsa.c
```diff
@@ -122,6 +227,66 @@ otcrypto_status_t otcrypto_rsa_public_key_construct(
...
 
+otcrypto_status_t otcrypto_rsa_public_key_deconstruct(
+    otcrypto_unblinded_key_t *public_key, otcrypto_word32_buf_t modulus,
+    uint32_t *exponent) {
+  if (public_key == NULL || public_key->key == NULL || modulus.data == NULL ||
+      exponent == NULL) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  // Entropy complex must be initialized for `hardened_memcpy`.
+  HARDENED_TRY(entropy_complex_check());
```
jadephilipoom: Since this function is for public keys, `hardened_memcpy` might be overkill; it's a lot slower than a normal copy and a lot of its defenses are to prevent side-channel leakage rather than only faults. I think it's fine to leave it in, but I'd maybe add a clarifying comment that it's for fault resistance just so future readers understand why it's there. I do think extra care with RSA public keys matters, because it's so easy to fault them into weak moduli.

sw/device/lib/crypto/impl/rsa.c
```diff
@@ -263,6 +423,200 @@ otcrypto_status_t otcrypto_rsa_private_key_from_exponents(
...
+      hardened_memcpy(d_p.data, sk->d_p.data, d_p.len);
+      hardened_memcpy(d_q.data, sk->d_q.data, d_q.len);
+      hardened_memcpy(i_q.data, sk->i_q.data, i_q.len);
+      return OTCRYPTO_OK;
+    }
+    default: {
+      // Invalid key size. Since the size was inferred, should be unreachable.
+      HARDENED_TRAP();
+      return OTCRYPTO_FATAL_ERR;
+    }
```
jadephilipoom: nit: the functions above have both this trap case at the end of the switch _and_ a trap case at the end of the function. Either version is okay with me (it's basically a minor code size vs defensive coding tradeoff) but I think it would be good to stay consistent.

sw/otbn/crypto/rsa_keygen.s
```diff
@@ -123,7 +126,511 @@ rsa_keygen:
...
+ * @param[out] dmem[rsa_n..rsa_n+(plen*2*32)] RSA public key modulus (n)
+       check value
+ * @param[in]  dmem[rsa_d_p..rsa_d_p+(plen*32)] first RSA private key private
+       exponent CRT component (d_p) check value
+ * @param[in]  dmem[rsa_d_q..rsa_d_q+(plen*32)] second RSA private key private
+       exponent CRT component (d_q) check value
+ * @param[in]  dmem[rsa_i_q..rsa_i_q+(plen*32)] RSA private key CRT
+       reconstruction coefficient (i_q) check value
+ * @param[in]  dmem[rsa_e..rsa_e+32] RSA private key private exponent (d)
+       check value
```
jadephilipoom:
```suggestion
 * @param[out]  dmem[rsa_d_p..rsa_d_p+(plen*32)] first RSA private key private
       exponent CRT component (d_p) check value
 * @param[out]  dmem[rsa_d_q..rsa_d_q+(plen*32)] second RSA private key private
       exponent CRT component (d_q) check value
 * @param[out]  dmem[rsa_i_q..rsa_i_q+(plen*32)] RSA private key CRT
       reconstruction coefficient (i_q) check value
 * @param[out]  dmem[rsa_e..rsa_e+32] RSA private key private exponent (d)
       check value
```

sw/otbn/crypto/rsa_keygen.s
```diff
@@ -123,7 +126,511 @@ rsa_keygen:
...
+ * private key by checking the following values:
+ *
+ *   - The n check value (in rsa_n) will be computed as p * q and should equal
+ *     the public modulus (n).
+ *   - The d_p check value (in rsa_d_p) will be computed as e * d_p mod (p - 1)
+ *     and should equal 1.
+ *   - The d_q check value (in rsa_d_q) will be computed as e * d_q mod (q - 1)
+ *     and should equal 1.
+ *   - The i_q check value (in rsa_i_q) will be computed as q * i_q mod p and
+ *     should equal 1.
```
jadephilipoom: At some point for better FI resistance we should make these multi-bit values like `kHardenedBoolTrue`. Doesn't have to block this PR, but would you mind making an issue for it? One interesting way to do it that would also provide some side-channel resistance is to generate a random mask, then multiply the mask in with the multiplication that should result in 1, and return both the mask and the multiplication result (which should be equal).

jadephilipoom (approved):
Whew, thanks for the hard work on this! Looks good to me, just a few minor comments.

As future work, I think it would be good to make tests that cover _all_ the reasons a private key might be invalid (`d_p`, `d_q`, `i_q`, `n`, `d`, `p`, and `q`) just so we're making sure that every boolean condition is getting checked (it seems too easy otherwise to accidentally leave one out of the list). But that's not blocking.

pqcfox:
Perfect, that should address the issues; will let CI run through though as I did add the extra `HARDENED_TRAP()` in addressing one comment.

pqcfox:
Added #169 and #170 to track the two mentioned issues.

PR #171, opened 2025-12-16
[crypto] Convert private key check API to return OTCRYPTO_ASYNC_INCOMPLETE

pqcfox:
This PR quickly converts the private key import and check API from a blocking API to a truly asynchronous one, returning `OTCRYPTO_ASYNC_COMPLETE` if the OTBN key checks are still running.

### DIFF
--- a/sw/device/lib/crypto/impl/rsa/rsa_keygen.c
+++ b/sw/device/lib/crypto/impl/rsa/rsa_keygen.c
@@ -370,6 +370,9 @@ status_t rsa_key_check_2048_finalize(const rsa_2048_public_key_t *public_key,
                                      const rsa_2048_private_key_t *private_key,
                                      hardened_bool_t check_primes,
                                      hardened_bool_t *key_valid) {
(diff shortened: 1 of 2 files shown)


sw/device/lib/crypto/impl/rsa/rsa_keygen.c
```diff
@@ -370,6 +370,9 @@ status_t rsa_key_check_2048_finalize(const rsa_2048_public_key_t *public_key,
                                      const rsa_2048_private_key_t *private_key,
                                      hardened_bool_t check_primes,
                                      hardened_bool_t *key_valid) {
+  // Return `OTCRYTPO_ASYNC_INCOMPLETE` if OTBN not done.
+  HARDENED_TRY(otbn_assert_idle());
+
   // Spin here waiting for OTBN to complete.
   OTBN_WIPE_IF_ERROR(otbn_busy_wait_for_done());
```
jadephilipoom: These can probably be deleted now in all cases, right?

PR #184, opened 2026-01-13
[crypto] ML-DSA (FIPS 204) signature implementation

mkannwischer:
This commit adds an integration of mldsa-native into OpenTitan, following the same integration pattern as ML-KEM. mldsa-native is a high-speed high-assurance C90 implementation of ML-DSA (FIPS 204); mldsa-native also has native code for AArch64 and x86_64, but that is unrelated to this integration.

The C code in mldsa-native has been proved to be memory-safe (no buffer overflow) and type-safe (no integer overflow) using the C Bounded Model Checker (CBMC), with the exception of the small portion of code paths enabled by `MLD_CONFIG_REDUCE_RAM` - we are working to remedy this, see below. C code is hardened against compiler-introduced timing side-channels through suitable barriers and constant-time patterns.

mldsa-native is currently in alpha release stage: production-ready, but some parts of the API may still change. Feedback is welcome. For more information, see https://github.com/pq-code-package/mldsa-native

Integration is achieved through a bazel dependency using mldsa-native's multi-level monobuild. A wrapper using OT cryptolib conventions is added. Tests are added comparing against known test vectors for the deterministic (derand) APIs for all three ML-DSA parameter sets (ML-DSA-44, ML-DSA-65, ML-DSA-87).

Similar to ML-KEM, the integration leverages OpenTitan's KMAC unit for massive performance gains. This required setting
`MLD_CONFIG_SERIAL_FIPS202_ONLY` to ensure only one SHAKE context is active at any time, as the KMAC accelerator holds only a single state.

Due to OpenTitan's memory constraints, this integration enables `MLD_CONFIG_REDUCE_RAM`, which streams the matrix A during signing operations instead of expanding it in full. This significantly impacts signing performance compared to the default configuration. The configuration option also uses unions to reduce overall memory footprint. CBMC proofs for code paths enabled by `MLD_CONFIG_REDUCE_RAM` are still facing some issues and will be completed at a later point (tracked at https://github.com/pq-code-package/mldsa-native/issues/826).

Stack Management:

To address OpenTitan's 16KB stack limit, this integration implements a custom bump allocator for ML-DSA's internal memory allocations. The allocator is enabled via `MLD_CONFIG_CUSTOM_ALLOC_FREE` and configured through `MLD_CONFIG_CONTEXT_PARAMETER`.

The bump allocator uses a simple linear allocation strategy with bounds checking. A context structure tracks the base pointer, total size, and current offset into a caller-provided work buffer. Allocations increment the offset; frees decrement it. If an allocation would exceed the buffer size, NULL is returned, triggering mldsa-native's built-in error handling. With correctly sized buffers, this should never occur in practice.

All ML-DSA API functions accept work buffer parameters sized according to the operation (keypair, sign, or verify). Buffer sizes are verified at compile time using static assertions against mldsa-native's `MLD_TOTAL_ALLOC_*` constants. The test suite allocates a single static work buffer sized to the maximum of all operations (62688 bytes for ML-DSA-87), resulting in peak stack usage well within the 16KB limit.

This approach moves the bulk of ML-DSA's internal allocations from stack to caller-provided buffers. A few kilobytes of allocations remain on the stack and will be addressed in future work (tracked at https://github.com/pq-code-package/mldsa-native/issues/867).

Benchmarking:

Benchmarking ML-DSA signature generation is challenging due to rejection sampling, which results in variable execution time. We cannot afford to benchmark enough signatures to obtain a statistical estimate, particularly when benchmarking using verilator. To estimate expected performance, the tests use the following trick. They benchmark signing with a message that requires 1 iteration in the rejection sampling loop, then benchmark signing with a message that requires 2 iterations in the rejection sampling loop, and linearly extrapolate to FIPS 204 expected iteration counts: 4.25 iterations for ML-DSA-44, 5.1 iterations for ML-DSA-65, and 3.85 iterations for ML-DSA-87. Note that this is an approximation as there are multiple rejection conditions with varying iteration costs. The benchmark mitigates this by selecting messages where rejection occurs due to the z-norm check, which accounts for the vast majority of rejections.

Hash ML-DSA mode is not yet integrated, but provisions are made in the API (`otcrypto_mldsa_sign_mode_t` enum) to add this in a follow-up PR.

### DIFF
--- a/sw/device/lib/crypto/impl/mldsa.c
+++ b/sw/device/lib/crypto/impl/mldsa.c
@@ -1,0 +1,590 @@
+// Copyright The mldsa-native project authors
+// Copyright zeroRISC Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/crypto/include/mldsa.h"
+
+#include <string.h>
+
+#include "sw/device/lib/base/math.h"
+#include "sw/device/lib/crypto/drivers/entropy.h"
+#include "sw/device/lib/crypto/impl/integrity.h"
+#include "sw/device/lib/crypto/impl/keyblob.h"
+#include "sw/device/lib/crypto/impl/mldsa/mldsa_native_monobuild.h"
+#include "sw/device/lib/crypto/impl/status.h"
+
+// Module ID for status codes.
+#define MODULE_ID MAKE_MODULE_ID('m', 'l', 'd')
+
+// Static assertions to verify buffer sizes match mldsa-native
+_Static_assert(kOtcryptoMldsa44WorkBufferKeypairWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_44_KEYPAIR,
+               "ML-DSA-44 keypair work buffer size mismatch");
+_Static_assert(kOtcryptoMldsa44WorkBufferSignWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_44_SIGN,
+               "ML-DSA-44 sign work buffer size mismatch");
+_Static_assert(kOtcryptoMldsa44WorkBufferVerifyWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_44_VERIFY,
+               "ML-DSA-44 verify work buffer size mismatch");
+
+_Static_assert(kOtcryptoMldsa65WorkBufferKeypairWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_65_KEYPAIR,
+               "ML-DSA-65 keypair work buffer size mismatch");
+_Static_assert(kOtcryptoMldsa65WorkBufferSignWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_65_SIGN,
+               "ML-DSA-65 sign work buffer size mismatch");
+_Static_assert(kOtcryptoMldsa65WorkBufferVerifyWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_65_VERIFY,
+               "ML-DSA-65 verify work buffer size mismatch");
+
+_Static_assert(kOtcryptoMldsa87WorkBufferKeypairWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_87_KEYPAIR,
+               "ML-DSA-87 keypair work buffer size mismatch");
+_Static_assert(kOtcryptoMldsa87WorkBufferSignWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_87_SIGN,
+               "ML-DSA-87 sign work buffer size mismatch");
+_Static_assert(kOtcryptoMldsa87WorkBufferVerifyWords * sizeof(uint32_t) ==
+                   MLD_TOTAL_ALLOC_87_VERIFY,
+               "ML-DSA-87 verify work buffer size mismatch");
+
+otcrypto_status_t otcrypto_mldsa44_keygen(
+    otcrypto_unblinded_key_t *public_key, otcrypto_blinded_key_t *secret_key,
+    uint32_t work[kOtcryptoMldsa44WorkBufferKeypairWords]) {
+  HARDENED_TRY(entropy_complex_check());
+
+  uint32_t seed[ceil_div(kOtcryptoMldsa44SeedBytes, sizeof(uint32_t))];
+  HARDENED_TRY(entropy_csrng_instantiate(
+      /*disable_trng_input=*/kHardenedBoolFalse, &kEntropyEmptySeed));
+  HARDENED_TRY(entropy_csrng_generate(&kEntropyEmptySeed, seed, ARRAYSIZE(seed),
+                                      /*fips_check=*/kHardenedBoolTrue));
+  HARDENED_TRY(entropy_csrng_uninstantiate());
+
+  otcrypto_const_byte_buf_t seed_buf = {.data = (unsigned char *)seed,
+                                        .len = sizeof(seed)};
+  return otcrypto_mldsa44_keypair_derand(seed_buf, public_key, secret_key,
+                                         work);
+}
+
+otcrypto_status_t otcrypto_mldsa44_keypair_derand(
+    otcrypto_const_byte_buf_t seed, otcrypto_unblinded_key_t *public_key,
+    otcrypto_blinded_key_t *secret_key,
+    uint32_t work[kOtcryptoMldsa44WorkBufferKeypairWords]) {
+  if (seed.len != kOtcryptoMldsa44SeedBytes) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  if (public_key == NULL || secret_key == NULL) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  if (public_key->key_mode != kOtcryptoKeyModeMldsa44) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  if (secret_key->config.key_mode != kOtcryptoKeyModeMldsa44) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  if (public_key->key_length != kOtcryptoMldsa44PublicKeyBytes) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  if (secret_key->config.key_length != kOtcryptoMldsa44SecretKeyBytes) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+
+  // Write the unmasked secret key into the first share of the keyblob.
+  uint32_t *sk_share0;
+  uint32_t *sk_share1;
+  HARDENED_TRY(keyblob_to_shares(secret_key, &sk_share0, &sk_share1));
+  memset(sk_share1, 0, kOtcryptoMldsa44SecretKeyBytes);
+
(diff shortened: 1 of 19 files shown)


mkannwischer:
I'm still running benchmarks and will post the results once it completes.

@hanno-becker, could you take a look in the meantime, please?

sw/device/lib/crypto/impl/mldsa.c
```diff
@@ -0,0 +1,627 @@
...
+                         .size_words = kOtcryptoMldsa44WorkBufferSignWords,
+                         .offset_words = 0};
+
+  // Unmask the key for the underlying implementation.
+  uint32_t *sk = (uint32_t *)mld_alloc(&ctx, kOtcryptoMldsa44SecretKeyBytes);
+  if (sk == NULL) {
+    return OTCRYPTO_FATAL_ERR;
+  }
+  size_t sk_words = ceil_div(kOtcryptoMldsa44SecretKeyBytes, sizeof(uint32_t));
+  status_t status = keyblob_key_unmask(secret_key, sk_words, sk);
```
mkannwischer: @jadephilipoom, I'm not very sure about this one. 
The unmasking of the secret key is adopted from what you have done for mlkem-native. 
However, for ML-DSA this is much more costly in terms of memory (ML-DSA secret keys are 4896 bytes). 

If secret keys are generated by mldsa-native, then this unmasking is unnecessary as the second share is going to be zero anyway. If we can just use the first share directly we can save up to 4896 bytes of memory. 

I don't understand enough of Otcrypto conventions to make the call here.

sw/device/lib/crypto/impl/mldsa/mldsa_native_alloc.c
```diff
@@ -0,0 +1,27 @@
...
+  return ptr;
+}
+
+void mld_free(void *ptr, mld_alloc_ctx_t *ctx, size_t size_bytes) {
+  if (ptr == NULL) {
+    return;  // No-op if NULL
+  }
+
+  size_t size_words = MLD_ALIGN_UP(size_bytes) / sizeof(uint32_t);
+  ctx->offset_words -= size_words;
```
hanno-becker: SImilarly pedantic comment: One could check that `size_bytes` is word-aligned and below the current watermark (this would catch free'ing in non-reverse order, for example).

sw/device/lib/crypto/impl/mldsa/mldsa_native_monobuild.h
```diff
@@ -0,0 +1,30 @@
+// Copyright The mldsa-native project authors
+// Copyright zeroRISC Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef OPENTITAN_SW_DEVICE_LIB_CRYPTO_IMPL_MLDSA_MLDSA_NATIVE_MONOBUILD_H_
+#define OPENTITAN_SW_DEVICE_LIB_CRYPTO_IMPL_MLDSA_MLDSA_NATIVE_MONOBUILD_H_
+
+#define MLD_CONFIG_NO_SUPERCOP
```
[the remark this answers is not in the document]
mkannwischer: Good point. That is an artifact from when `mldsa_native.h` was not yet including the config itself.
I'll move it.

sw/device/lib/crypto/impl/mldsa/mldsa_native_alloc.c
mkannwischer: Added a check that `ptr` is as expected (`base + offset - size`) - trapping on mismatch.

mkannwischer:
Out of interest I also ran benchmarks **without using KMAC** (i.e., using the default Keccak C implementation in mldsa-native - which admitedly isn't very suitable for 32-bit platforms). This is like 2.6 - 3.6x slower than with KMAC (factors in parentheses). 


| Operation | ML-DSA-44 | ML-DSA-65 | ML-DSA-87 |
|-----------|-----------|-----------|-----------|
| **KeyGen** | 2,423,815 cycles (3.0x) | 4,188,959 cycles (3.3x) | 6,955,985 cycles (3.6x) |
| **Sign** | 3,627,516 (1 iteration) (1.9x)<br>6,125,624 (2 iterations) (2.2x)<br>11,746,367 (4.25 iterations) (2.6x) | 5,685,894 (1 iteration) (2.0x)<br>9,765,770 (2 iterations) (2.4x)<br>22,413,385 (5.1 iterations) (2.9x) | 9,093,296 (1 iteration) (2.3x)<br>15,991,753 (2 iterations) (2.7x)<br>28,753,898 (3.85 iterations) (3.1x) |
| **Verify** | 2,510,390 cycles (2.6x) | 4,162,759 cycles (2.8x) | 7,029,168 cycles (3.2x) |


raw low: 
```
I00001 test_rom.c:190] kChipInfo: scm_revision=54697461
I00002 test_rom.c:267] Test ROM complete, jumping to flash (addr: 20000480)!
I00001 ottf_main.c:168] Running sw/device/tests/crypto/mldsa_functest.c
I00002 mldsa_functest.c:163] Generating keypair...
I00003 profile.c:22] otcrypto_mldsa44_keypair_derand took 2423815 cycles or 24 ms @ 100 MHz.
I00004 mldsa_functest.c:169] Signing...
I00005 mldsa_functest.c:191] 2-iteration message signing time: 6125624 cycles
I00006 mldsa_functest.c:208] 1-iteration message signing time: 3627516 cycles
I00007 mldsa_functest.c:216] Expected signing time (4.25 iterations): 11746367 cycles
I00008 mldsa_functest.c:219] Verifying...
I00009 profile.c:22] otcrypto_mldsa44_verify took 2510390 cycles or 25 ms @ 100 MHz.
I00010 mldsa_functest.c:264] Generating keypair...
I00011 profile.c:22] otcrypto_mldsa65_keypair_derand took 4188959 cycles or 41 ms @ 100 MHz.
I00012 mldsa_functest.c:270] Signing...
I00013 mldsa_functest.c:292] 2-iteration message signing time: 9765770 cycles
I00014 mldsa_functest.c:309] 1-iteration message signing time: 5685894 cycles
I00015 mldsa_functest.c:317] Expected signing time (5.1 iterations): 22413385 cycles
I00016 mldsa_functest.c:320] Verifying...
I00017 profile.c:22] otcrypto_mldsa65_verify took 4162759 cycles or 41 ms @ 100 MHz.
I00018 mldsa_functest.c:365] Generating keypair...
I00019 profile.c:22] otcrypto_mldsa87_keypair_derand took 6955985 cycles or 69 ms @ 100 MHz.
I00020 mldsa_functest.c:371] Signing...
I00021 mldsa_functest.c:393] 2-iteration message signing time: 15991753 cycles
I00022 mldsa_functest.c:410] 1-iteration message signing time: 9093296 cycles
I00023 mldsa_functest.c:418] Expected signing time (3.85 iterations): 28753898 cycles
I00024 mldsa_functest.c:421] Verifying...
I00025 profile.c:22] otcrypto_mldsa87_verify took 7029168 cycles or 70 ms @ 100 MHz.
I00026 ottf_main.c:110] Finished sw/device/tests/crypto/mldsa_functest.c
I00027 status.c:37] PASS!
[2026-01-13T13:22:10Z INFO  opentitantool::command::console] ExitSuccess("PASS!\r\n")
```

sw/device/lib/crypto/include/mldsa.h
```diff
@@ -0,0 +1,376 @@
...
+ */
+OT_WARN_UNUSED_RESULT
+otcrypto_status_t otcrypto_mldsa44_keygen(
+    otcrypto_unblinded_key_t *public_key, otcrypto_blinded_key_t *secret_key,
+    uint32_t work[kOtcryptoMldsa44WorkBufferKeypairWords]);
+
+/**
+ * ML-DSA-44 deterministic keypair generation.
+ *
+ * Generates a public/secret key pair from a given seed (for testing).
```
jadephilipoom: If it's intended for testing only, it probably should not be in this file. However, my impression was that deterministic keypair generation also has legitimate uses like restoring a keypair from a small seed if you're storage-space constrained. If that sounds right to you, I'd just remove the "(for testing)" parenthetical.

sw/device/lib/crypto/impl/mldsa/fips202_glue.h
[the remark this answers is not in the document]
mkannwischer: Unfortunately, not :( the functions needed are not the same.
We want to consolidate that at some point.

sw/device/lib/crypto/impl/mldsa.c
mkannwischer: okay cool, I have made that change.

sw/device/lib/crypto/include/mldsa.h
jadephilipoom: >     * Is there a way to mark the functions static here, but still expose them to the tests? (I think the answer is no - if it would exist I would propose to use that for now and add additional APIs once it is clearer what APIs we need).

No, not that I know of.

>     * Do we want to support the deterministic mode? If yes, is the current implementation of `_derand` with an all-zero input acceptable?

Yes, I think we do want to support deterministic mode (including e.g. for generating ML-KEM keys from key manager derived seeds).

>     * Do we need to support the external-mu mode?

I think it would be good to support it.

>     * Do we want the call the API `_derand` (aligned with ML-KEM) or `_internal` (aligned with FIPS204)

I prefer `_derand` since we're exposing the API.

jadephilipoom:
The CI failure looks unrelated to me. I re-ran the test in case it's just being flaky (there were a few other tests in the same job that failed once and then succeeded on the second attempt). ROM_EXT should have nothing to do with this, though.

PR #217, opened 2026-03-17
[ci,acc] Adds support for ACC PQC smoke test in Verilator and adds it to CI

apinise:
This PR restructures the previous configuration for passing the PQC parameter from a simulation config to the DV environment.
- Removes the use of environment variable and instead adds a -CFLAG to configure the `iss_wrapper.cc` at build time which will create the `accsim` ISS model.
- Verilator pre_dv fusesoc core now accepts a `--flag pqc` to configure the ACC instantiation and appropriate core dependencies/-CFLAG options.

It also updates the `smoke_pqc` directory to use the appropriate fusesoc invocation and integrates into CI
- Updated the Verilator waiver to prevent warnings with KMAC after instantiating into the testbench. Warnings consisted of `UNUSED` for signals in the top-level testbench used for visibility, `UNOPTFLAT` for combinational assignments to struct fields that rely on the value of other fields in the same struct, and `VARHIDDEN` for having multiple FSM in KMAC that share state names in their enum declarations.
- Updated the expected `smoke_pqc/smoke_test.s` results based on model behavior after adding a KMAC transaction and the vector instructions.
- Added a new CI test to run the ACC PQC smoke test on verilator with the vector ISA hardware.


LouisTheLuis (approved):
LGTM.

PR #229, opened 2026-03-27
[cryptotest] Add Ed25519 tests using ACVP test vectors 

mkannwischer:
This PR adds ACVP-based Ed25519 tests. As of now the cryptolib only implements HashEd25519.
ACVP has testvectors for both pure Ed25519 and HashEd25519, while Wycheproof only has pure Ed25519 vectors.

The ACVP testvectors revealed two minor bugs regarding the context strings that the cryptolib could handle; those are fixed in separate commits.
See individual commit messages for the details.

### DIFF
--- a/sw/device/lib/crypto/impl/ecc/ed25519.c
+++ b/sw/device/lib/crypto/impl/ecc/ed25519.c
@@ -4,6 +4,7 @@
 
 #include "sw/device/lib/crypto/impl/ecc/ed25519.h"
 
+#include "sw/device/lib/base/math.h"
 #include "sw/device/lib/crypto/drivers/entropy.h"
 #include "sw/device/lib/crypto/drivers/rv_core_ibex.h"
 #include "sw/device/lib/crypto/impl/ecc/ed25519_insn_counts.h"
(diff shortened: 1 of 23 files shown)


sw/device/lib/crypto/impl/ecc/ed25519.c
pqcfox: [nit] from prior it looks like `context_length`'s @param was missing, could be worth adding here to explicitly state that it's a byte length now?

jadephilipoom (approved):
LGTM, thanks Matthias!

pqcfox (approved):
Looks excellent! Left a few small notes.

sw/device/lib/crypto/impl/ecc/ed25519.c
pqcfox: (dup, apologies!)

sw/host/cryptotest/testvectors/parsers/nist_acvp_eddsa_parser.py
```diff
@@ -0,0 +1,140 @@
...
+        if not group.get("preHash", False):
+            continue
+
+        d = list(bytes.fromhex(group["d"]))
+        q = list(bytes.fromhex(group["q"]))
+        for test in group["tests"]:
+            msg_bytes = bytes.fromhex(test["message"])
+            prehash = hashlib.sha512(msg_bytes).digest()
+            ctx_hex = test.get("context", "")
+            context = list(bytes.fromhex(ctx_hex)) if ctx_hex else []
```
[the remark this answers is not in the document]
mkannwischer: Actually, 
```py
>>> bool("")
False
```

But also,
```py
>>> list(bytes.fromhex(""))
[]
```

So the if is not needed indeed. Will fix.

sw/device/lib/crypto/impl/ecc/ed25519.c
mkannwischer: fixed.

PR #230, opened 2026-03-27
[crypto] Update `otcrypto_interface.h` and backport map file analysis 

mkannwischer:
Our  `otcrypto_interface.h` is outdated (e.g., it was not updated in https://github.com/zerorisc/expo/pull/204) - `otcrypto_export_size`, hence, fails to compile.
This fixes that and also backports a changes to allow cryptolib map file analysis which as a side effect also causes this test to be picked up by our CI (at least that is what I hope - testing now).

### DIFF
--- a/sw/device/silicon_owner/bare_metal/BUILD
+++ b/sw/device/silicon_owner/bare_metal/BUILD
@@ -9,6 +9,11 @@ load("//rules/opentitan:defs.bzl", "fpga_params", "opentitan_binary", "opentitan
 
 package(default_visibility = ["//visibility:public"])
 
+filegroup(
+    name = "bare_metal_start",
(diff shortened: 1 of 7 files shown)


pqcfox:
#240 should fix the Verilated EG failure; since it's unrelated to this PR in any case, feel free not to wait for that to go through.

pqcfox (approved):
Looks excellent, thank you so much for fixing this!

mkannwischer:
> Looks excellent, thank you so much for fixing this!

Thanks! Rebased and it failed because the cryptolib diverged further. Fixed in https://github.com/zerorisc/expo/compare/e264dea3a63e2d0c65a361bda458061bac446c87..8df02643a0b3f74d8f7f3a40cbfd31421d991d4d

PR #231, opened 2026-03-31
[acc] Add clobbered-register annotation checker

mkannwischer:
Add a static checker that verifies clobbered-register docstring
annotations against the actual registers written, as determined by
the existing information-flow analysis.

The check currently only covers a part of the routines
(those that are currently supported by the information-flow
analysis script).

Also adjusted the docstrings for the covered routines - most of those were not aligned with the output of the information-flow script, but some of those differences are just cosmetic (like changing `x2, x3` to `x2 to x3`)

### DIFF
--- a/sw/acc/crypto/tests/BUILD
+++ b/sw/acc/crypto/tests/BUILD
@@ -3,9 +3,140 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-load("//rules:acc.bzl", "acc_consttime_test", "acc_library", "acc_sim_test", "acc_sim_test_suite")
+load("//rules:acc.bzl", "acc_clobbered_regs_test", "acc_consttime_test", "acc_library", "acc_sim_test", "acc_sim_test_suite")
 
 package(default_visibility = ["//visibility:public"])
+
+# Clobbered-register checks: verify that docstring annotations match
+# the actual registers written (determined by information-flow analysis).
+# N-dependent RSA/bignum helpers, routines taking a WDR index in a GPR,
+# and sha3_shake (no acc_binary) are omitted.
+#
+# TODO: investigate why these take 30s-200s; tagged "manual" until fixed.
+_SLOW_CLOBBERED_REGS_SUBS = [
+    ("run_p384", "p384_base_mult"),
+    ("run_p384", "p384_base_mult_checked"),
+    ("run_p384", "scalar_mult_int_p384"),
+    ("run_p384", "p384_scalar_mult"),
+    ("run_p384", "p384_sign"),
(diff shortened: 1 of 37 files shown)


mkannwischer:
@pqcfox, @jadephilipoom, do you think something like this would be useful? Or am I overengineering this?

sw/acc/crypto/tests/BUILD
```diff
@@ -3,10 +3,141 @@
...
+load("//rules:acc.bzl", "acc_clobbered_regs_test", "acc_consttime_test", "acc_library", "acc_sim_test", "acc_sim_test_suite")
 
 package(default_visibility = ["//visibility:public"])
 
+# Clobbered-register checks: verify that docstring annotations match
+# the actual registers written (determined by information-flow analysis).
+# N-dependent RSA/bignum helpers, routines taking a WDR index in a GPR,
+# and sha3_shake (no acc_binary) are omitted.
+#
+# TODO: investigate why these take 30s-200s; tagged "manual" until fixed.
```
jadephilipoom, line 15: I think this is probably because for `loopi` instructions the information-flow backend unrolls loops, and these ones all include the P384 scalarmult loop repeated 448 times. If so, we could fix it by making the backend treat large loops the same way it treats loops with a non-constant number of iterations, but at the cost of potentially missing some details in large loops that depend on a specific number of iterations, and perhaps making the behavior a little less predictable.

sw/acc/crypto/tests/BUILD
```diff
@@ -3,10 +3,141 @@
...
+    ("run_p384", "p384_verify"),
+]
+
+[acc_clobbered_regs_test(
+    name = elf + "_" + sub + "_clobbered",
+    srcs = ["//sw/acc/crypto:" + src],
+    subroutine = sub,
+    tags = ["manual"] if (elf, sub) in _SLOW_CLOBBERED_REGS_SUBS else [],
+    deps = ["//sw/acc/crypto:" + elf],
+) for sub, src, elf in [
```
jadephilipoom, line 31: Hmm, rather than this big loop it would be nice to run this almost as a linter. Could we put it under the top-level `quality` maybe with the other linters, and automatically run it on everything that's not specifically excluded?

jadephilipoom (approved):
Thanks Matthias, looks good to me! I think this would be nice to have.

sw/acc/crypto/tests/BUILD
jadephilipoom: As of https://github.com/zerorisc/expo/pull/247 this issue is fixed; all the consttime checks take less than 5s.

sw/acc/crypto/tests/BUILD
mkannwischer: I will try this when i find some time.

PR #233, opened 2026-04-01
[crypto] Add pure Ed25519 signing and verification

mkannwischer:
- Depends on #229 

This PR adds pure Ed25519 signature generation and verification as described in RFC 8032.

Verification is a straightforward extension only touching the C-side as the hash is computed using HMAC prior to invoking ACC - this is fine as all processed data is public. 

Signing is more involved as both required hashes are computed in ACC. To be able to support arbitrary message lengths, we have to stream in the message twice resulting in a 5-step API: INIT/UPDATE/MID/UPDATE/FINAL. I tried to share as much code as possible with the pre-hash mode.

This allows us to cover more ACVP test vectors and also support the Wycheproof testvectors.

I have also extended the functest accordingly and cleaned up a few typos on the way.

See the individual commit messages for details.

### DIFF
--- a/sw/acc/crypto/run_ed25519.s
+++ b/sw/acc/crypto/run_ed25519.s
@@ -6,15 +6,19 @@
  * Entrypoint for Ed25519 operations.
  *
  * This binary has the following modes of operation:
- * 1. ED25519_MODE_SIGN: generate an Ed25519 signature using caller-provided secret key
- * 2. ED25519_MODE_VERIFY: verify an Ed25519 signature
+ * 1. ED25519_MODE_SIGN_PREHASH: generate a HashEd25519 signature (single-shot)
+ * 2. ED25519_MODE_SIGN_PURE_INIT: start incremental pure Ed25519 signing
+ * 3. ED25519_MODE_SIGN_PURE_UPDATE: feed more message data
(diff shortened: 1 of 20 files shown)


sw/acc/crypto/run_ed25519.s
```diff
@@ -6,15 +6,19 @@
  * Entrypoint for Ed25519 operations.
  *
  * This binary has the following modes of operation:
- * 1. ED25519_MODE_SIGN: generate an Ed25519 signature using caller-provided secret key
- * 2. ED25519_MODE_VERIFY: verify an Ed25519 signature
+ * 1. ED25519_MODE_SIGN_HASH: generate a HashEd25519 signature (single-shot)
```
jadephilipoom: nit: I might prefer `ED25519_MODE_SIGN_PREHASH` for this mode name, since there's hashing that occurs as part of the pure Ed25519 operation and it might be slightly confusing, while "prehash" is unambiguous.

jadephilipoom (approved):
LGTM, thanks for this!

sw/acc/crypto/run_ed25519.s
mkannwischer: Agreed. Fixed.

PR #235, opened 2026-04-07
[bench] Rearrange benchmarking directory to be ACC-specific.

jadephilipoom:
Previously, the ACC benchmarking scripts `collect_stats.sh` and `analyze_stats.py` lived in a top-level `bench/` directory of their own, which made a lot less sense once they were moved out of the ACC-focused `expo-otbn-pqc` development repo. The first commit moves these scripts to `bench/acc`. I considered instead moving them to `sw/acc/bench`, but after discussion with @pqcfox it seems like it might be a good idea to have a top-level `bench` directory anyway that goes beyond ACC.

The second commit is just a quick little fixup so the scripts by default dump the logs into `bench/acc/logs` instead of into the same directory as the scripts. The previous version resulted in a pretty messy `git status` if you run a lot of benchmarks, not that I would know anything about that :innocent:


pqcfox (approved):
Looks great, thanks so much!

PR #236, opened 2026-04-09
[crypto] Fix cryptotest test grouping by algorithm and update `cryptolib_test_suite`

pqcfox:
This chore PR contains two small changes:

- One (coarsely) groups the tests in `sw/device/tests/crypto/BUILD` by algorithm and general type, e.g. symmetric/asymmetric. (Over a long time, the ordering of tests started to scatter a bit.)
- One updates `cryptolib_test_suite` to include all tests (some were previously missing) in alphabetical order to pass the Bazel lint.

As a sanity check, the first commit has been verified to not drop any tests with the following one-liner, which should be reproducible locally:

```
 diff <(git show master:sw/device/tests/crypto/BUILD | grep "name =" | sort) <(git show kat/cryptolib-test-suite-update:sw/device/tests/crypto/BUILD | grep "name = " | sort)
 ```

Running this should yield no output, e.g. that no rules were added or removed. One can verify that the one-liner would otherwise detect some difference by removing the `| sort` from each subprocess.


pqcfox:
Reviewing myself to make sure the first commit includes everything correctly to make sure all's clear, then will mark as ready for review.

pqcfox:
For review, I'd highly suggest locally using `git diff --histogram`; this captures the diff for the first commit much more cleanly.

mkannwischer (approved):
Thanks! LGTM. I spot checked that the first commit only shuffles around tests and that the second commit includes all tests.

PR #237, opened 2026-04-09
[acc] Add specialized point doubling routine for P-384.

jadephilipoom:
Adds point doubling and a test. The routine is currently unused, but will later help speed up the core constant-time scalar multiplication routine as well as signature verification. The `proj_add_test` completes in a total of 1496 cycles, while `proj_double_test` completes in 965 cycles (and can take its input entirely in registers instead of loading and storing from memory). Since scalar multiplication is a double-and-add loop at its core, I think the extra code size for the specialized routine is well spent.


sw/acc/crypto/tests/p384_proj_double_test.s
```diff
@@ -0,0 +1,79 @@
...
+.word 0xb5dbd0ba
+.word 0x760dc196
+.word 0x15e0be5c
+.word 0x6d678c86
+.word 0x65ca4081
+.word 0xde946124
+.word 0x3c315786
+.word 0xb0a1d256
+.word 0x0c1eea5a
+.zero 8
```
mkannwischer: Why 8 bytes? Shouldn't this be 16? Doesn't matter because of the `balign` of course.

sw/acc/crypto/p384_base.s
```diff
@@ -909,6 +909,203 @@ proj_add_p384:
 
   ret
 
+
+/**
+ * P-384 point doubling in projective space.
+ *
+ * Adds a (valid) point to itself in-place.
+ *
+ * Uses the "dbl-2007-bl" formula from the Explicit Formulas Database:
```
mkannwischer: This is in fact dbl-2007-bl-2, right? (as the link below suggests).

mkannwischer (approved):
Thanks. Two nits, otherwise this looks great for me.

I checked that the code indeed corresponds to dbl-2007-bl-2 and confirmed the tests pass and match the cycle counts listed. 

Is it worth to include the cycle count in the commit message for future reference? Are there any conventions for that?

One tests is very little, but once this is plugged into the higher level code, I suppose we do have decent test coverage including the special handling of the point at infinity, do we?

sw/acc/crypto/tests/p384_proj_double_test.s
jadephilipoom: Yes, it should be 16! Fixed now (the `.zero` is just there for clarity but still nice if it's accurate)

sw/acc/crypto/p384_base.s
jadephilipoom: Yes, fixed, thanks!

jadephilipoom:
> Is it worth to include the cycle count in the commit message for future reference? Are there any conventions for that?

The second commit in #238 has the cycle counts after doubling is integrated (as well as some other changes, these were all a bit entangled because of register pressure). There aren't concrete conventions for it but I often include them if there's a big change in a top-level operation.

> One tests is very little, but once this is plugged into the higher level code, I suppose we do have decent test coverage including the special handling of the point at infinity, do we?

Yes, once it's integrated into the P-384 scalarmult it gets covered by cryptotest. The special point at infinity handling is always hit for scalarmult, because the accumulator starts at the origin and then gets doubled immediately; if you remove the special handling every use of scalarmult will produce incorrect results, so we already get test failures on the ACC simulator tests before even running wycheproof on it.

jadephilipoom:
Since this code is not called by anything tested on Verilator and the Verilator tests have been flaky on other PRs (also passed on #238 which incorporates this change), I think the CI failure is spurious. I'll merge anyway if no one objects by the end of my workday.

pqcfox:
Currently working on a fix for the Verilated EG tests; confirming the failure is spurious, feel free to merge any time.

PR #240, opened 2026-04-09
[verilator] Add shutdown request mechanism to prevent early UART closes in CI

pqcfox:
This PR introduces a new shutdown request mechanism for Verilator tests to prevent spurious test failures from early UART closures.

Note that this PR only implements this change for the EarlGrey top, as this is the only top Verilator tests support at the moment. Analogous changes can be made later to `chip_sim_tb.sv` for other tops as they become supported.

Prior to this PR, Verilator tests would sometimes fail due to the chip simulation testbench calling `$finish` before `opentitantool console` could drain the UART buffer, triggering a test failure even after success was visibly reported over UART. To fix this, a 'virtual' `REQUEST_EXIT` pin has been added to the GPIO DPI to trigger a graceful shutdown of the verilated testbench, with the host then waiting a fixed timeout for the child subprocess to exit before killing it as before. The handshake goes as follows:

- OTTF running in Verilator emits a `PASS!` over the Verilator UART DPI interface
- opentitantool picks up on this PASS as usual, registers the test result, and drops the `Verilator` context struct
- This in turn triggers a call to `Verilator::shutdown()`, which _now_ 
  - Raises the virtual `REQUEST_EXIT` pin via the file descriptor shared with the GPIO DPI 
  - Calls `shutdown_with_timeout` on the Verilator subprocess struct, `Subprocess`
- The Verilated EG binary sees `REQUEST_EXIT` raised and calls `$finish` from within `gpiodpi.sv`
- The `Subprocess` struct in `shutdown_with_timeout()` waits for the Verilated EG process to terminate
  - If a timeout expires, then `shutdown_with_timeout()` kills the child process as previous

This handshake ensures that opentitantool is no longer listening on the UART DPI file descriptor by the time Verilator calls `$finish`.

This PR introduces a dependency on `wait-timeout`, which is a Rust crate allowing one process to wait on a child process to terminate with a timeout. License for `wait-timeout` is standard MIT/Apache 2.0.


third_party/rust/Cargo.toml
```diff
@@ -84,6 +85,7 @@ thiserror = "2"
 tiny-keccak = {version = "2.0.2", features = ["cshake"]}
 tokio = { version = "1", features = ["full"] }
 typetag = "0.2"
+wait-timeout = "0.2.1"
```
matutem, line 88: Is this a version requirement on a dependency? I don't see what "wait-timeout" indicates, was it a pre-existing module/package/thing? It seems is not created in this PR.

pqcfox: It is, it's the most recent version of an external package from crates.io which I had to introduce to properly handle waiting on the Verilator subprocess to close with a timeout (so that I could send `kill` and print a warning should the subprocess not respond to the `REQUEST_EXIT` pin being asserted in a reasonable amount of time).

sw/host/ot_transports/verilator/src/transport.rs
```diff
@@ -23,7 +24,14 @@ use opentitanlib::util::parse_int::ParseInt;
 use super::gpio::{GpioInner, VerilatorGpioPin};
 use super::subprocess::{Options, Subprocess};
 
+/// Baud for UART communication with host.
 const UART_BAUD: u32 = 40;
+/// Timeout in milliseconds to wait after requesting exit for the verilator subprocess to finish.
+const REQUEST_EXIT_TIMEOUT: u64 = 2000;
+/// 'Virtual' pin used for requesting the verilator subprocess exit gracefully.
+const REQUEST_EXIT_PIN: u8 = 254;
```
rtorok-zr, line 32: Just to confirm, these values should be okay to keep hard-coded as long as no top has more than 254 GPIOs?

rtorok-zr (approved):
LGTM, incredible work troubleshooting this Kat!

sw/host/ot_transports/verilator/src/transport.rs
pqcfox: Yup! I figured this was a safe assumption, but glad you brought it up explicitly. Since the Verilator harnesses break out only the necessary pins for simulation, and the imaginary pin 255 was already reserved for all tops, I figured running out of pins was unlikely to happen.

PR #242, opened 2026-04-12
[kmac,rtl] Adds masking support for the sideload AppIntf in KMAC

apinise:
This PR extends the masking in AppIntf masking through the entirety of KMAC.

- The struct `app_req_t` in `kmac_pkg.sv` now contains `data_share0` and `data_share1`.
- `kmac_app.sv` now processes two shares in masking mode for the ACC connection through its `AppCfgDynamic` mode. Other connections, including ROM, LC, and KeyMgr still use unmasked data path and the second share is driven to '0 in the respective block levels.
- `kmac_msg.sv` has a second msgfifo in masked mode to hold the second share before forwarding it to the `sha3` module. The FIFO status are tied together for error handling and fault detection.
- `kmac_app_host_seq.sv` now drives two shares to the ACC app interface when in masked mode, other interfaces remain '0.
- `kmac_scoreboard.sv` will unmask the input message to send to the DPI-C model.


LouisTheLuis (approved):
LGTM.

PR #245, opened 2026-04-24
[acc] Streamine trigger_fault routines and better recognize no-ops in information flow.

jadephilipoom:
This PR:
- moves the `trigger_fault_if_fg0_z` and `trigger_fault_if_fg0_not_z` to separate files so P-256, P-384, and RSA keygen do not all maintain separate copies of this logic
- adjusts the information-flow backend rules to better detect certain special cases where operands being the same changes the information flow (`bn.xor`, `bn.sub`, `bn.movr`)
- changes the name of the complement routine to `trigger_fault_if_not_fg0_z`, because it makes more sense to me than `fg0_not_z`
- changes the mechanism for producing a fault to `bn.movr` with an invalid register instead of `bn.lid` with an invalid offset, for a few reasons:
  -  gives an `ILLEGAL_INSN` error instead of a `BAD_DATA_ADDR` area (I think that's a little bit clearer for debugging)
  - does not depend on DMEM size
  - plays more nicely with the information-flow checker given the no-op detection for `bn.movr`, so we don't have to exclude/ignore these paths
  - saves a few cycles (the new routines are 4 cycles each, compared to 8 each before)

### DIFF
--- a/sw/acc/crypto/boot.s
+++ b/sw/acc/crypto/boot.s
@@ -1,4 +1,5 @@
 /* Copyright lowRISC contributors (OpenTitan project). */
+/* Copyright zeroRISC Inc. */
 /* Licensed under the Apache License, Version 2.0, see LICENSE for details. */
 /* SPDX-License-Identifier: Apache-2.0 */
 
@@ -168,7 +169,7 @@ attestation_keygen:
      The check fails if both sides are not equal.
      FG0.Z <= (y^2) mod p == (x^2 + ax + b) mod p */
   bn.cmp   w18, w19
-  jal      x1, trigger_fault_if_fg0_not_z
+  jal      x1, trigger_fault_if_not_fg0_z
 
   ecall
 
--- a/sw/acc/crypto/rsa_keygen.s
+++ b/sw/acc/crypto/rsa_keygen.s
@@ -1333,20 +1333,12 @@ modinv_f4:
   bn.addi    w23, w31, 1
   bn.cmp     w22, w23
 
-  /* Get the FG0.Z flag into a register.
-       x2 <= CSRs[FG0] & 8 = FG0.Z << 3 */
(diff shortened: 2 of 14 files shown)


sw/acc/crypto/rsa_keygen.s
mkannwischer: Can this be dropped?

sw/acc/crypto/trigger_fault_if_fg0_z.s
```diff
@@ -0,0 +1,33 @@
...
+ * clobbered flag groups: none
+ */
+.globl trigger_fault_if_fg0_z
+trigger_fault_if_fg0_z:
+  /* Read the FG0.Z flag (position 3).
+       x2 <= FG0.Z << 3 */
+  csrrw     x2, FG0, x0
+  andi      x2, x2, 8
+
+  /* Causes an ILLEGAL_INSN error if x2 == 1; otherwise just does a no-op move
```
mkannwischer: x2 can be 0 or 8 here, never 1.

sw/acc/crypto/boot.s
```diff
@@ -1,3 +1,7 @@
+/* Copyright zeroRISC Inc. */
+/* Licensed under the Apache License, Version 2.0, see LICENSE for details. */
+/* SPDX-License-Identifier: Apache-2.0 */
+
```
mkannwischer: Please remove the duplicated license/SPDX header.

mkannwischer (commented):
Thanks @jadephilipoom. Looks good to me overall.
I left a couple of nits.

sw/acc/crypto/rsa_keygen.s
jadephilipoom: Yes, thanks!

mkannwischer (approved):
Thanks @jadephilipoom for the update. LGTM.

PR #246, opened 2026-04-24
[acc] Minor improvements to information-flow checker.

jadephilipoom:
On top of #245, draft until that PR is merged.

These were some small improvements I made along the way towards a bigger extension to the information-flow checker that I'll post as a follow-up PR; I separated these out for easier review. There's more detail in the commit messages, but in summary:
- print out control-flow nodes more nicely for debugging
- automatically resolve branches where both sides of the comparison are known constants
- allow specifying _public_ values instead of _secret_ values for constant-time checks, which makes it easier to ensure nothing has been forgotten

### DIFF
--- a/hw/ip/acc/util/check_const_time.py
+++ b/hw/ip/acc/util/check_const_time.py
@@ -12,6 +12,7 @@ from shared.control_flow import program_control_graph, subroutine_control_graph
 from shared.decode import decode_elf
 from shared.information_flow_analysis import (get_program_iflow,
                                               get_subroutine_iflow,
+                                              parse_information_flow_node,
                                               stringify_control_deps)
 
 
@@ -43,15 +44,26 @@ def main() -> int:
               'in the form "reg:value", e.g. x3:5. Only GPRs are accepted as '
               'required constants.'))
     parser.add_argument(
-        '--secrets',
+        '--secret',
         nargs='+',
         type=str,
         required=False,
         help=('Initial secret information-flow nodes. If not provided, '
-              'assume everything is secret; check that the subroutine or '
-              'program has only one possible control-flow path regardless '
-              'of input.'))
+              'assume everything is secret unless --public is specified; '
+              'check that the subroutine or program has only one possible '
+              'control-flow path regardless of input.'))
+    parser.add_argument(
+        '--public',
+        nargs='+',
+        type=str,
+        required=False,
+        help=('Initially NOT secret information-flow nodes. If specified, '
+              'assume everything else is secret. Cannot be specified at the '
+              'same time as --secret.'))
     args = parser.parse_args()
+
+    # Load the program.
+    program = decode_elf(args.elf)
 
     # Parse initial constants.
     if args.constants is None:
@@ -62,6 +74,21 @@ def main() -> int:
                              'program; use --subroutine to analyze a specific '
                              'subroutine.')
         constants = parse_required_constants(args.constants)
+
+    if args.secret and args.public:
+        raise ValueError('Cannot specify --secret and --public together.')
+
+    # Parse the secrets as information-flow nodes.
+    secret_nodes = set()
+    if args.secret:
+        for secret in args.secret:
+            secret_nodes.add(parse_information_flow_node(secret))
+
+    # Parse the public values as information-flow nodes.
+    public_nodes = set()
+    if args.public:
+        for public in args.public:
+            public_nodes.add(parse_information_flow_node(public))
 
     # Compute control graph and get all nodes that influence control flow.
     program = decode_elf(args.elf)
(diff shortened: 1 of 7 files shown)


hw/ip/acc/util/analyze_information_flow.py
```diff
@@ -10,7 +10,8 @@
 from shared.control_flow import program_control_graph, subroutine_control_graph
 from shared.decode import decode_elf
 from shared.information_flow import InformationFlowGraph
-from shared.information_flow_analysis import (get_program_iflow,
+from shared.information_flow_analysis import (get_dmem_symbols,
```
mkannwischer: I'm getting a 
`ImportError: cannot import name 'get_dmem_symbols' from 'shared.information_flow_analysis' ` here.

hw/ip/acc/util/shared/information_flow_analysis.py
```diff
@@ -661,3 +686,24 @@ def stringify_control_deps(program: ACCProgram,
             pc_strings.append('{} at PC {:#x}'.format(insn.mnemonic, pc))
         out.append('{} (via {})'.format(node, ', '.join(pc_strings)))
     return out
+
+
+def parse_information_flow_node(program: ACCProgram, x: str) -> str:
```
mkannwischer: `program` is not used in this function - is that intentional?

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -15,8 +15,29 @@
 from .operand import Operand, RegOperandType
 
 FLAG_NAMES = ['c', 'm', 'l', 'z']
-SPECIAL_REG_NAMES = ['mod', 'acc', 'acch']
-READONLY = ['x0']
+
+# Deliberately invalid CSR address used by unimp.
+UNIMP_CSR_ADDR = 3072
```
mkannwischer: This is unused? Was that intentional?

hw/ip/acc/util/analyze_information_flow.py
mkannwischer: ~Given that CI did not flag this, aren't we running the constant-time tests in CI?~
Given that CI did not flag this, should we extend the CI so this script is exercised? Remind me what the main purpose of the information-flow check is?

hw/ip/acc/util/analyze_information_flow.py
```diff
@@ -82,7 +86,7 @@ def main() -> int:
             raise ValueError('Cannot require initial constants for a whole '
                              'program; use --subroutine to analyze a specific '
                              'subroutine.')
-        constants = parse_required_constants(args.constants)
+        constants = parse_required_constants(args.constants, dmem_symbols)
```
mkannwischer: I think you didn't change `parse_required_constants`? It only takes 1 argument.

mkannwischer (changes requested):
Thanks @jadephilipoom. I think you forgot to push something or something went wrong when rebasing.

I'm a little concerned that this still made it through CI even do it seems to break the information flow analysis. ~Are we not running the constant time tests in CI on every PR? Shouldn't we?~  The constant-time tests are not depending on this script - I got that wrong.

hw/ip/acc/util/analyze_information_flow.py
jadephilipoom: We run the  `check_const_time.py` script in CI; this `analyze_information_flow.py` script uses the same backend but gives more detail. It's run manually if you want to understand what's going on in the code better.

hw/ip/acc/util/shared/information_flow_analysis.py
jadephilipoom: This is an artifact of detangling these changes from #247; in #247 it gets used. I'll remove it here.

hw/ip/acc/util/shared/information_flow.py
jadephilipoom: Again an artifact from #247, it gets used there to detect `unimp` calls and not error out on the invalid CSR. Will remove here.

hw/ip/acc/util/analyze_information_flow.py
jadephilipoom: Once again a rebase artifact, should be fixed now! Sorry for all the interference, these two were quite intertwined but since #247 is pretty large I wanted to separate out as much as I could.

hw/ip/acc/util/check_const_time.py
```diff
@@ -43,16 +44,27 @@ def main() -> int:
...
+        nargs='+',
+        type=str,
+        required=False,
+        help=('Initially NOT secret information-flow nodes. If specified, '
+              'assume everything else is secret. Cannot be specified at the '
+              'same time as --secret.'))
     args = parser.parse_args()
 
+    # Load the program.
+    program = decode_elf(args.elf)
```
mkannwischer, line 66: There is another `decode_elf` in line 94. Can one be removed?

hw/ip/acc/util/check_const_time.py
```diff
@@ -63,6 +75,21 @@ def main() -> int:
                              'subroutine.')
         constants = parse_required_constants(args.constants)
 
+    if args.secret and args.public:
+        raise ValueError('Cannot specify --secret and --public together.')
+
+    # Parse the secrets as information-flow nodes.
+    secret_nodes = set()
```
mkannwischer, line 82: `secret_nodes` is never used, but I guess you  will need it later? feel free to ignore.

hw/ip/acc/util/shared/constants.py
```diff
@@ -180,11 +180,15 @@ def parse_required_constants(constants: List[str]) -> Dict[str, int]:
             raise ValueError(
                 f'Cannot parse required constant {token}: {reg} is not a '
                 'valid GPR name.')
-        if not value.isnumeric():
+        try:
+            if value.startswith('0x'):
+                value = int(value, 16)
+            else:
+                value = int(value)
```
mkannwischer, line 187: nit: Could be `value = int(value, 0)`?

mkannwischer (approved):
Thanks for addressing my comments.
A couple of more nits - feel free to ignore. LGTM.

PR #248, opened 2026-04-27
[ci,cryptolib] Add linting for Doxygen comments

mkannwischer:
This PR adds linting for the Doxygen comments to CI. Currently I have scoped it only to the cryptolib.

I fixed around 150 mistakes in the Doxygen comments.


sw/device/lib/crypto/include/mlkem.h
```diff
@@ -54,8 +54,10 @@ enum {
  * blob for the private key should have a length of 2x
  * ceil(kOtCryptoMlkem512SecretKeybytes / sizeof(uint32_t)) = 816 words.
  *
- * @param[out] pk public key dest, len >= `kOtcryptoMlKemPublicKeyBytes`.
- * @param[out] sk secret key dest, unmasked len `kOtcryptoMlKemSecretKeyBytes`.
+ * @param[out] public_key public key dest, len >=
+ * `kOtcryptoMlKemPublicKeyBytes`.
```
jadephilipoom: nit: it would be nice to keep these on one line for easier visual scanning, and the names are a little self-documenting -- maybe something like:

```suggestion
 * @param[out] public_key dest, len >= `kOtcryptoMlKemPublicKeyBytes`.
```

I'd also be OK moving the length requirements to the comment above.

(same comment goes for a few other places in this file)

jadephilipoom (approved):
Nice to have this check actually run!

sw/device/lib/crypto/include/mlkem.h
mkannwischer: Yes, good point. Compressed all of them down to one line. Much prettier. Thanks

